PR #8, opened 2019-08-14
MultiPoint

visr:
From https://geojson.org/: GeoJSON supports the following geometry types: Point, LineString, Polygon, MultiPoint, MultiLineString, and MultiPolygon.

In this package, the only one of those not here is MultiPoint. Or did you leave it out on purpose?

### DIFF
--- a/src/basic_types.jl
+++ b/src/basic_types.jl
@@ -265,10 +265,9 @@ struct MultiPolygon{
         Element <: AbstractPolygon{Dim, T},
         A <: AbstractVector{Element}
     } <: AbstractVector{Element}
+
     polygons::A
 end
-
-
 
 function MultiPolygon(polygons::AbstractVector{P}; kw...) where P <: AbstractPolygon{Dim, T} where {Dim, T}
     MultiPolygon(meta(polygons; kw...))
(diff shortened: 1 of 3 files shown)


SimonDanisch:
MultiPoint is just a `AbstractVector{<: Point}` right?
I think I didn't really see the point to have that have its own type, but sure, why not, even if it's just for consistency ;)

visr:
Yeah I added it mainly for consistency.
The same thing can be said for the `MultiLineString` and `MultiPolygon` right? Not that that is an argument to have it. I'm still not sure whether it is better to use `Vector{Geom}` vs `MultiGeom`. At least you made the `MultiGeom` behave like `Vector{Geom}` ;)

We had that discussion before with @yeesian though, pasted below. Not sure if we made up our minds other than they are common?

> Simon Danisch [19:45]
> I'm wondering if we should just get rid off all these multi types^^... wouldn't `const MultiLineString = AbstractVector{LineString}` be enough? (edited) 

> Yeesian Ng [19:46]
> i guess it would. But people might want to treat it as a single object rather than as a vector of objects

> Simon Danisch [19:46]
> for what?
> I mean, I guess I can see that... Afterall the same could be said for Polygon
> so maybe we should just make it a subtype of `AbstractVector{LineString}`

> Yeesian Ng [19:48]
> https://sgillies.net/2010/03/24/origin-of-the-multi-geometry.html and https://sgillies.net/2014/01/02/maybe-the-shapefile-was-right-after-all.html (edited) 

> Simon Danisch [19:49]
> makes sense :wink:
> ok, let's just make them inherit then

> Martijn Visser [19:52]
> Yeah I guess we can reduce the amount of complexity considerably that way. I remember struggling with shapely (geos) because we didn't know ahead of time whether an operation would return a polygon or a multipolygon

> Martijn Visser [19:53]
> Need to keep in mind though that multipolygons are quite common, and are often used as a way to share metadata, e.g. one record is one district here: https://github.com/JuliaGeo/GeoInterface.jl/pull/12#issuecomment-355061122

> Simon Danisch [19:54]
> yeah, I agree :wink: so multipolys should be first class objects

> Yeesian Ng [19:57]
> the hope is that we’ll be sufficiently comprehensive in functionality (in julia) for both polygons and multipolygons such that it doesn’t become of concern to the user

> Martijn Visser [19:57]
> first class support, second class struct :wink:

> Yeesian Ng [19:58]
> lots of occasions where I see python code (a) only working for polygons or multipolygons, or (b) having if-statements to handle both

> Martijn Visser [20:00]
> yeah I image if we do a union of two polygons then we should return a vector of polygons, and this can contain either one or two depending if they intersected

SimonDanisch:
Yeah fair enough ;) I think it won't hurt, so merged!
We should just try to have our interface functions defined on `AbstractVector{<: AbstractPoint}`, which then can work for both!

src/basic_types.jl
```diff
@@ -283,9 +282,32 @@ struct MultiLineString{
...
+function MultiLineString(linestrings::AbstractVector{L}; kw...) where L <: AbstractVector{LineP{Dim, T, P}} where {Dim, T, P}
+    MultiLineString(meta(linestrings; kw...))
+end
+
+Base.getindex(ms::MultiLineString, i) = ms.linestrings[i]
+Base.size(ms::MultiLineString) = size(ms.linestrings)
+
+struct MultiPoint{
+    Dim, T <: Real,
+    Element <: Point{Dim, T},
```
SimonDanisch: Wait, that should be AbstractPoint!

Issue #10, opened 2019-08-14
Question about type parameters


visr:
If you take for instance the definition of `LineString`:
https://github.com/SimonDanisch/GeometryBasics.jl/blob/2a36f67effe9cc623ac82f09ce76c7c66007222c/src/basic_types.jl#L169-L175

I get that `Dim` and `T` should be there, representing the number of dimensions and type of the coordinates.

But `P`, the type of `Point`, and `V`, the the type of the vector of points, seem not needed. Why are they there? Do you need them to enforce something?

Also looking at these two:
https://github.com/SimonDanisch/GeometryBasics.jl/blob/2a36f67effe9cc623ac82f09ce76c7c66007222c/src/basic_types.jl#L133-L134
You use mainly `LineP`, but can't we do without?

Sorry if it's a bit basic but I'd like to get a good handle on this.

SimonDanisch:
If I remember correctly, you can't leave out the P in `<: LineP{Dim, T, P}` and `AbstractVector{LineP{Dim, T, P}} `.. If you do, it won't dispatch and type check correctly anymore :( 
It's annoying, but I don't think we can do without?

PR #12, opened 2019-08-14
remove LineP

visr:
See discussion in #10.
Does this have any downsides?
I tried to take the `V` out of `LineString` as well, but was not sure how to do that without restricting it's field to what was `V`. Or is there no need to restrict that field, since it is already taken care of by the subtyping? Am a bit confused.

### DIFF
--- a/src/basic_types.jl
+++ b/src/basic_types.jl
@@ -130,8 +130,7 @@ struct Simplex{
 end
 
 const NSimplex{N} = Simplex{Dim, T, N, P} where {Dim, T, P}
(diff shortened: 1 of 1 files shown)


SimonDanisch:
We should probably also have tests for type stability to see that we don't have any regressions!

visr:
Yeah how do we easily check that?

asinghvi17:
`@inferred` perhaps?

SimonDanisch:
Hm so parts of me like this, to reduce complexity & compilation times (possibly?), and another part thinks, oh no, now this is not strictly typed anymore :-O

visr:
Yeah I think I understand. I'll close the PR. Feel free to reopen if you still want to consider it more.

PR #18, opened 2020-03-07
Finish this up to work with everything

SimonDanisch:
@mkborregaard @sjkelly change of plans...
GeometryTypes contains so much code I don't really know what it does - or that I really want to get rid of, that it wasnt much fun to work with...
So new plan is again, to finish up GeometryBasics to be up for the task!
Its just so much work 😢


sjkelly:
Where can I help?

src/geometry_primitives.jl
```diff
@@ -0,0 +1,316 @@
...
+struct Particle{N, T} <: GeometryPrimitive{N, T}
+    position::Point{N, T}
+    velocity::Vec{N, T}
+end
+
+"""
+A `Cylinder` is a 2D rectangle or a 3D cylinder defined by its origin point,
+its extremity and a radius. `origin`, `extremity` and `r`, must be specified.
+"""
+struct Cylinder{N,T<: AbstractFloat} <: GeometryPrimitive{N,T}
```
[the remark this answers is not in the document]
sjkelly: I agree with `Real` here, which would allow support for Intervals in robust predicate filters.

mkborregaard:
Sounds like a good plan. Let's make some clear packages for the work. Is the design final yet though? Minimal as possible, types that mimic C types for interop? Is the type hierarchy clear? Sorry for the naive questions but I've lost touch a bit.

src/boundingboxes.jl
```diff
@@ -0,0 +1,39 @@
+function Rect(geometry::AbstractArray{<: Point{N, T}}) where {N,T}
+    return Rect{N,T}(geometry)
+end
+
+"""
+Construct a HyperRectangle enclosing all points.
+"""
+function (t::Type{Rect{N1, T1}})(
```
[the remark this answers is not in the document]
SimonDanisch: I haven't cleaned up this file...I think I'll just introduce a `boundingbox` function going forward, instead of misusing the Rect constructor

SimonDanisch:
Yay, first milestone:
```julia
julia> scatter(rand(4))
┌ Warning: Printing Scene as text. You see this because you haven't loaded any backend (GLMakie, CairoMakie, WGLMakie),
│ or you loaded GLMakie, but it didn't build correctly. In the latter case,
│ try `]build GLMakie` and watch out for any warnings.
└ @ AbstractPlotting ~/Beacon/dev/AbstractPlotting/src/display.jl:104
Scene (960px, 540px):
events:
    window_area: GeometryBasics.HyperRectangle{2,Int64}([0, 0], [0, 0])
    window_dpi: 100.0
    window_open: false
    mousebuttons: Set(AbstractPlotting.Mouse.Button[])
    mouseposition: (0.0, 0.0)
    mousedrag: notpressed
    scroll: (0.0, 0.0)
    keyboardbuttons: Set(AbstractPlotting.Keyboard.Button[])
    unicode_input: Char[]
    dropped_files: String[]
    hasfocus: false
    entered_window: false
plots:
   *Axis2D{...}
   *Scatter{...}
subscenes:
   *scene(960px, 540px)
```

SimonDanisch:
E.g @sjkelly we'll need to port the tests from GeometryTypes & add more ;) 
Of course you could also give feedback etc :)

SimonDanisch:
Next Milestone :) 
image

src/triangulation.jl
```diff
@@ -0,0 +1,163 @@
...
+            return false;
+        end
+    end
+    return true;
+end
+
+"""
+Triangulates a Polygon given as a `contour`::AbstractArray{Point} without holes.
+It will return a Vector{`facetype`}, defining indexes into `contour`
+"""
```
asinghvi17:
```suggestion
"""
    faces(contour::AbstractArray{Point}, [facetype = GLTriangleFace])

Triangulates a Polygon given as an `AbstractArray{Point}` without holes.
It will return a Vector{`facetype`}, defining indexes into `contour`
"""

File: test/runtests.jl [hunk 4]
```diff
@@ -1,270 +1,302 @@
...
-        polygon22 = Polygon(LineString(Point{2, Int}[(20, 35), (10, 30), (10, 10), (30, 5), (45, 20), (20, 35)]),
-            [LineString(Point{2, Int}[(30, 20), (20, 15), (20, 25), (30, 20)])])
-        multipolygon2 = MultiPolygon([polygon21, polygon22])
-        @test size(multipolygon2) === (2,)
-        @test multipolygon2[1] === polygon21
-        @test multipolygon2[2] === polygon22
     end
+end
 
+@testset "decompose/triangulateion" begin
```
asinghvi17:
```suggestion
@testset "decompose/triangulation" begin

**Comment by Tokazama** [comment]
Would it be helpful if I did any of these l?
- review and add to docs
- Add tests
- Add some needed functionality

**Comment by SimonDanisch** [comment]
Yes, extremely helpful :)

File: README.md [hunk 5]
```diff
@@ -6,7 +6,7 @@
 
 # GeometryBasics.jl
 
-Basic Geometry Types. 
+Basic Geometry Types.
 This package aimes to offer a standard set of Geometry types, which easily work with metadata, query frameworks on geometries and different memory layouts.
```
Tokazama: aimes -> aims

src/basic_types.jl
```diff
@@ -18,10 +19,22 @@ abstract type AbstractNgonFace{N, T} <: AbstractFace{N, T} end
 
 abstract type AbstractSimplex{Dim, N, T} <: StaticVector{Dim, T} end
 
-@fixed_vector Point AbstractPoint
-
 
+"""
+    coordinates(geometry)
+Returns the edges/vertices/coordinates of a geometry
+"""
```
Tokazama: Do you we want to include return types on these attributes accessors?
```
"""
    coordinates(geometry) -> AbstractPoint

Returns the edges/vertices/coordinates of a geometry
"""
```

README.md
asinghvi17:
```suggestion
This package aims to offer a standard set of geometry types, which work with metadata, query frameworks on geometries, and different memory layouts.

File: src/basic_types.jl (hunk 6 above)
Comment (asinghvi17) [reply src/basic_types.jl]: That would be nice, IMO.

**Comment by Tokazama** [comment]
I'm sure you're already thinking about this, but there's a ton here and it would be nice to get as much supported by other packages. Off the top of my head, would it be possible to use packages like CoordinatTransformations.jl and OffstArrays.jl? Maybe sharing developer burden with a broader community for some of these?

**Comment by asinghvi17** [comment]
CoordinateTransformations.jl already works with GeometryTypes, no reason it shouldn't also work with GeometryBasics since they share implementations for points, etc.  (CoordinateTransformations only expects an implementation of `getindex`, or some arraylike structure, and basic math definitions for its transformations).

File: src/offsetintegers.jl [hunk 7]
```diff
@@ -0,0 +1,62 @@
+
+"""
+OffsetInteger type mainly for indexing.
+* `O` - The offset relative to Julia arrays. This helps reduce copying when
+communicating with 0-indexed systems such ad OpenGL.
```
asinghvi17:
```suggestion
communicating with 0-indexed systems, such as OpenGL.

**Comment by Tokazama** [comment]
> CoordinateTransformations.jl already works with GeometryTypes, no reason it shouldn't also work with GeometryBasics since they share implementations for points, etc. (CoordinateTransformations only expects an implementation of getindex, or some arraylike structure, and basic math definitions for its transformations).

I was thinking of some code I saw with cylinder rotation. The other thing was that it would be convenient if rotations could be applied to an entire mesh, set of points, etc. instead of having to apply this manually or by broadcasting. Or would all of this be better implemented in another upstream package. I'm just thinking in terms of how nice it would be to have optimized rotations and projection routines readily available.

**Comment by asinghvi17** [comment]
I think we could overload arbitrary transformations for meshes - after all, to my knowledge you only need to transform and recalculate vertices and normals (if rotating).

**Comment by Tokazama** [comment]
> I think we could overload arbitrary transformations for meshes - after all, to my knowledge you only need to transform and recalculate vertices and normals (if rotating).

Exactly. It would be a very nice way to get an already well tested method of rotations built in. Should we put that off for another PR though? This is already getting pretty big and it seems like it would be easier to just take the current stuff for rotations out and implement it in a subsequent PR that can focus on that.

**Comment by asinghvi17** [comment]
Yes, we could base another PR off this branch.  It shouldn't be too hard - all transformations are supposed to subtype `Transformation`, so it would be fairly easy to overload all of those.

A prototype would look something like this:
```julia
function (t::CoordinateTransformations.Transformation)(m::T) where T <: AbstractMesh
    newpoints = t.(m.points)
    return T(newpoints, m.faces)
end
```

I'm not sure how the mesh interface works, so that last line will have to be reworked.  Still, you get the idea...

Tokazama:
After going through the documentation I was wondering why we have `contains` if it is exactly equivalent to `in`?

SimonDanisch:
Ah, we should remove that! I guess that sneaked in while copy & pasting code from GeometryTypes!

SimonDanisch:
This is almost finished :) If anyone wants to help testing, that'd be an amazing help!

PR #32, opened 2020-05-06
Bump minimum Julia to 1.2

sjkelly:
`hasproperty` is required and only 1.2+

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,6 @@ os:
 julia:
   - 1.2
   - 1
-  - nightly
-matrix:
-  allow_failures:
-    - julia: nightly
-  fast_finish: true
 notifications:
   email: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -3,7 +3,8 @@ language: julia
 os:
   - linux
 julia:
-  - 1.3
+  - 1.2
```
SimonDanisch, line 6: I stopped testing on any other version than the officially release julia version ;) 
So we should leave this at `1`

sjkelly: The latest is in the line below. It is good to test the minimum of the Project.toml. Otherwise downstream CI and users on random old versions will have breakage.

SimonDanisch: Lets go for a compromise^^

PR #36, opened 2020-05-10
fix rectangle normals

ffreyer:
I added extra coordinates to Rect3D. Now each face has its own set of coordinates, and normals act as surface normals.


ffreyer:
Btw, shouldn't `minimum` and `maximum` return the true minimum and maximum?
For example, with `r = Rect3D(Poitn3f0(0), Vec3f0(-1))` you currently get `minimum(r) = Vec3f0(0)`.

SimonDanisch:
Oh, good point :-O I guess, I kinda assume, you dont have negative widths... But then again, it's not enforced by the constructor!

ffreyer:
I think there are situations where it's convenient to use negative widths. Say you want to add a floor rectangle in some 3D plot. Then it's more convenient to write `Rect3D(Point3f0(0), Vec3f0(1, 1, -0.2))` than `Rect3D(Point3f0(0, 0, -0.2), Vec3f0(1, 1, 0.2))` imo.

SimonDanisch:
It is, but we may want to rewrite that in the constructor to still end up with positive widths, I suppose

PR #39, opened 2020-05-16
add a few tests on Rect and Line

visr:
Commits best viewed separately, since the last one is just whitespace.
Hoping this will get coverage over 50% ;)


SimonDanisch:
Awesome :) :)

PR #58, opened 2020-06-02
Polytype -> Polytope

blegat:
By the way, it seems `Polytope{N}` denotes a polytope in `N` dimension, not a polytope made of `N` points since the first type argument is named `Dim`.


SimonDanisch:
Thanks! Almost forgot about this PR :)

Issue #63, opened 2020-06-14
GB.Meta() errors for Multi Geometries


Sov-trotter:
For Geometries like MultiPoint, MultiPolygon the following error occurs, 
 ```
julia> using GeometryBasicsjulia> p1 = Point(3, 1)
2-element Point{2,Int64} with indices SOneTo(2):
 3
 1
julia> poi = meta(p1, city="Abuja", rainfall=1221.2)
2-element PointMeta{2,Int64,Point{2,Int64},(:city, :rainfall),Tuple{String,Float64}} with indices SOneTo(2):
 3
 1
julia> MultiPoint([p1])
1-element MultiPoint{2,Int64,Point{2,Int64},Array{Point{2,Int64},1}}:
 [3, 1]
julia> meta(MultiPoint([p1]), city="Abuja", rainfall=1221.2)
ERROR: Metadata needs to be an array with the same length as data items. Found: String
```
also passing the metadata as a NamedTuple reproduces the error, 
 ```
julia> using GeometryBasics; const GB = GeometryBasics
GeometryBasics

julia> tup = (featurecla = "Land", min_zoom = 0, scalerank = 0)
(featurecla = "Land", min_zoom = 0, scalerank = 0)

julia> polys = [Polygon(rand(Point{2, Float32}, 20)) for i in 1:10];

julia> multipoly = MultiPolygon(polys; tup...)  
ERROR: Metadata needs to be an array with the same length as data items. Found: String
 ```

Sov-trotter:
One possible workaround is the way this issue was solved https://github.com/JuliaGeometry/GeometryBasics.jl/issues/52, 
but as @visr pointed out the meta() call is going [here](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/master/src/metadata.jl#L115-L119) rather than going to the [overload](https://github.com/JuliaGeometry/GeometryBasics.jl/blob/master/src/metadata.jl#L106-L109) method.

visr:
A key point here is that generally a multi geometry is interpreted similarly to a single geometry, i.e. one feature, or one row in a table.

It may be fine to have multi geometries implemented as vectors of their single part versions, but they are not the same thing. In a vector of points, each point can have it's own metadata. In a MultiPoint, there is only one geometry, so only one metadata. Having a single multi geometry as a StructArray therefore does not quite seem the right approach.

PR #65, opened 2020-06-16
Add metatype for MultiPolygon and LineString

Sov-trotter:
#63


Sov-trotter:
With the tests and readme updated I think this PR Is ready to go?

PR #68, opened 2020-07-07
Remove type piracy in `isnan`

yha:
Ref JuliaGeometry/GeometryTypes.jl#207.

### DIFF
--- a/src/rectangles.jl
+++ b/src/rectangles.jl
@@ -294,14 +294,14 @@ function Base.to_indices(A::AbstractMatrix{T}, I::Tuple{Rect2D{IT}}) where {T, I
 end
 
 function minmax(p::StaticVector, vmin, vmax)
-    isnan(p) && return (vmin, vmax)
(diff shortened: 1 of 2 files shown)


src/rectangles.jl
```diff
@@ -294,14 +294,14 @@ function Base.to_indices(A::AbstractMatrix{T}, I::Tuple{Rect2D{IT}}) where {T, I
...
-    isnan(p) && return (vmin, vmax)
+    any(isnan, p) && return (vmin, vmax)
     min.(p, vmin), max.(p, vmax)
 end
 
 # Annoying special case for view(Vector{Point}, Vector{Face})
 function minmax(tup::Tuple, vmin, vmax)
     for p in tup
-        isnan(p) && continue
+        any(isnan, p) && continue
```
yha, line 304: Not sure if this change is necessary (because I don't understand how this method is supposed to be called), but I don't think it hurts either.

Issue #69, opened 2020-07-11
coordinates of a Polygon


visr:
I have a pretty simple question but I can't seem to find the answer. I'll walk through my attempts, and have some more general questions about `coordinates`.

*Given `pol`, how do I get a vector like `pts` back? I only need the exterior.*

```julia
pts = Point{2, Int}[(3, 1), (4, 4), (2, 4), (1, 2), (3, 1)]
pol = Polygon(pts)
```

My first guess was `coordinates`
```julia
julia> coordinates(pol.exterior)
4-element reinterpret(GeometryBasics.Ngon{2,Int64,2,Point{2,Int64}}, ::TupleView{Tuple{Point{2,Int64},Point{2,Int64}},2,1,Array{Point{2,Int64},1}}):
 Line([3, 1] => [4, 4])
 Line([4, 4] => [2, 4])
 Line([2, 4] => [1, 2])
 Line([1, 2] => [3, 1])
```

Somewhat surprisingly to me `coordinates` returns lines, not points. Of course I can dig deeper and get the coordinates of those:

```julia
julia> coordinates.(coordinates(pol.exterior))
4-element Array{StaticArrays.SArray{Tuple{2},Point{2,Int64},1,2},1}:
 [[3, 1], [4, 4]]
 [[4, 4], [2, 4]]
 [[2, 4], [1, 2]]
 [[1, 2], [3, 1]]
```

But now every inner point appears twice, which is not what I want. I think the `TupleView` is giving a lines view over the points that I want, but how do I get them? The `coordinates` documentation says to try `decompose`:

>     coordinates(geometry)
> Returns the edges/vertices/coordinates of a geometry. Is allowed to return lazy iterators!
Use `decompose(ConcretePointType, geometry)` to get `Vector{ConcretePointType}` with
`ConcretePointType` to be something like `Point{3, Float32}`.

But that errors, is this a bug or am I using it wrongly?

```julia
julia> decompose(Point{2, Int}, pol.exterior)
ERROR: MethodError: no method matching Int64(::Point{2,Int64})
Closest candidates are:
  Int64(::Union{Bool, Int32, Int64, UInt32, UInt64, UInt8, Int128, Int16, Int8, UInt128, UInt16}) at boot.jl:708
  Int64(::Ptr) at boot.jl:718
  Int64(::Float32) at float.jl:706
  ...
Stacktrace:
[...]

PR #73, opened 2020-07-26
Give out points of geometries

Sov-trotter:
PR for #69  
So this overloads the `convert_simplex()` method to break down a linestring. 
```julia
ls1=LineString([Point(1, 2), Point(4,5), Point(10, 8), Point(1, 2)])
decompose(Point{2, Int}, ls1)
6-element Array{Point{2,Int64},1}:
 [1, 2]
 [4, 5]
 [4, 5]
 [10, 8]
 [10, 8]
 [1, 2]
```
The repetition of Points is the only visible issue currently. Expected is: 
```julia
ls1.points.parent.data
4-element Array{Point{2,Int64},1}:
 [1, 2]
 [4, 5]
 [10, 8]
 [1, 2]
```

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -116,10 +116,9 @@ function decompose(::Type{P}, pol::Polygon) where {P<:AbstractPoint}
     if isempty(pol.interiors)
         return decompose(P, pol.exterior)
     else
-        arr = P[]
-        push!(arr, decompose(P, pol.exterior)...)
+        arr = copy(decompose(P, pol.exterior))
         for i in pol.interiors
-            push!(arr, decompose(P, i)...)
+            append!(arr, decompose(P, i))
         end
         return arr
     end
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -116,11 +116,15 @@ function decompose(::Type{P}, pol::Polygon) where {P<:AbstractPoint}
     if isempty(pol.interiors)
         return decompose(P, pol.exterior)
     else
-        points = Array{Union{Point, Array}}[]
-        push!(points, decompose(P, pol.exterior), [decompose(P, pol.interiors[1]) for i in pol.interiors])
-        return points
+        arr = P[]
+        push!(arr, decompose(P, pol.exterior)...)
+        for i in pol.interiors
+            push!(arr, decompose(P, i)...)
+        end
+        return arr
     end
 end
+
 decompose(::Type{P}, ls::LineString) where {P<:AbstractPoint} = ls.points.parent.data
 decompose_uv(primitive) = decompose(UV(), primitive)
 decompose_uvw(primitive) = decompose(UVW(), primitive)
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -111,11 +111,8 @@ end
 function decompose(::Type{T}, primitive) where {T}
     return collect_with_eltype(T, primitive)
 end
-"""
-decompose a linesting into points
-"""
-decompose(::Type{P}, ls::LineString) where {P<:AbstractPoint} = ls.points.parent.data
 
+decompose(::Type{P}, ls::LineString) where {P<:AbstractPoint} = ls.points.parent.data
 decompose_uv(primitive) = decompose(UV(), primitive)
 decompose_uvw(primitive) = decompose(UVW(), primitive)
 decompose_normals(primitive) = decompose(Normal(), primitive)
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -112,6 +112,15 @@ function decompose(::Type{T}, primitive) where {T}
     return collect_with_eltype(T, primitive)
 end
 
+function decompose(::Type{P}, pol::Polygon) where {P<:AbstractPoint}
+    if isempty(pol.interiors)
+        return decompose(P, pol.exterior)
+    else
+        points = Array{Union{Point, Array}}[]
+        push!(points, decompose(P, pol.exterior), [decompose(P, pol.interiors[1]) for i in pol.interiors])
+        return points
+    end
+end
 decompose(::Type{P}, ls::LineString) where {P<:AbstractPoint} = ls.points.parent.data
 decompose_uv(primitive) = decompose(UV(), primitive)
 decompose_uvw(primitive) = decompose(UVW(), primitive)
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -111,6 +111,10 @@ end
 function decompose(::Type{T}, primitive) where {T}
     return collect_with_eltype(T, primitive)
 end
+"""
+decompose a linesting into points
+"""
+decompose(::Type{P}, ls::LineString) where {P<:AbstractPoint} = ls.points.parent.data
 
 decompose_uv(primitive) = decompose(UV(), primitive)
 decompose_uvw(primitive) = decompose(UVW(), primitive)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -373,7 +373,21 @@ end
     @test coordinates(m) === points
 
     linestring = LineString(Point{2, Int}[(10, 10), (20, 20), (10, 40)])
-    @test decompose(Point{2, Int}, linestring) == Point{2, Int64}[[10, 10], [20, 20], [10, 40]]
+    pts = Point{2, Int}[(10, 10), (20, 20), (10, 40)]
+    linestring = LineString(pts)
+    pts_decomp = decompose(Point{2, Int}, linestring)
+    @test pts === pts_decomp
+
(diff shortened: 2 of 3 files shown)


Sov-trotter:
Now the thing is `convert_simplex` receives only `Line` so it's pretty hard to avoid the repetitions. We can either introduce a new `check_repeat()` method in the `collect_with_eltype()` or maybe overload it for `LineString/polygon/other higher geometries` specifically?

SimonDanisch:
How about overloading decompose?

Sov-trotter:
Ah. That's doable. BTW can we use the `points.parent.data` to return our points? Or use the method in the latest commit and iterate over it to remove the duplicates?

Sov-trotter:
If things are right, this can be merged?

src/interfaces.jl
```diff
@@ -112,6 +112,15 @@ function decompose(::Type{T}, primitive) where {T}
     return collect_with_eltype(T, primitive)
 end
 
+function decompose(::Type{P}, pol::Polygon) where {P<:AbstractPoint}
+    if isempty(pol.interiors)
+        return decompose(P, pol.exterior)
+    else
+        points = Array{Union{Point, Array}}[]
```
SimonDanisch, line 119: `decompose(::Type{P<:AbstractPoint}, x)` always needs to return `Vector{P <: AbstractPoint}`

Sov-trotter:
So the new overload is for giving out points of a polygon. 
A polygon has two parts, and exterior and interiors. In the latest commit, if there are no interiors, an array of points of exterior is given out, but it interiors are present, the data is given out as:
```julia
pol = Polygon(LineString(Point{2, Int}[(5, 1), (3, 3), (4, 8), (1, 2), (5, 1)]), [LineString(Point{2, Int}[(2, 2), (3, 8),(5, 6) , (3, 4), (2, 2)]), LineString(Point{2, Int}[(3, 2), (4, 5),(6, 1) , (1, 4), (3, 2)])])

decompose(Point{2, Int}, pol)2-element Array{Array{Union{Array, Point},N} where N,1}:
 [[5, 1], [3, 3], [4, 8], [1, 2], [5, 1]]
 [Point{2,Int64}[[2, 2], [3, 8], [5, 6], [3, 4], [2, 2]], Point{2,Int64}[[2, 2], [3, 8], [5, 6], [3, 4], [2, 2]]]
```
Is this nested Array approach better? Or what might be a better way to have interiors spread out?

src/interfaces.jl
Sov-trotter: So what can be the right way incase we have multiple interiors?

visr: The two options I can think of is either make one long `Vector{P <: AbstractPoint}` that holds both exterior and interior, or throw an error. But perhaps the former is better since if people want only the points that make up the exterior. They can decompose only `pol.exterior`.

test/runtests.jl
```diff
@@ -371,6 +371,9 @@ end
     points = decompose(Point2f0, Circle(Point2f0(0), 1))
     m = GeometryBasics.mesh(points)
     @test coordinates(m) === points
+
+    linestring = LineString(Point{2, Int}[(10, 10), (20, 20), (10, 40)])
+    @test decompose(Point{2, Int}, linestring) == Point{2, Int64}[[10, 10], [20, 20], [10, 40]]
```
visr, line 376: This is good, but we can even express it more strongly:

```julia
pts = Point{2, Int}[(10, 10), (20, 20), (10, 40)]
linestring = LineString(pts)
pts_decomp = decompose(Point{2, Int}, linestring)
@test pts === pts_decomp
```
Meaning, the result of the decomposition are the input points.

Could you add a test for the polygon decompose as well, with and without interiors?

src/interfaces.jl
Sov-trotter: But the only issue here being, how will we differentiate between the interiors then?

visr: We can't. But it's impossible to both retain that and return a vector of points.

Sov-trotter: So there is also a discussion going on here https://github.com/JuliaPlots/AbstractPlotting.jl/pull/479#discussion_r467369773

Do you think we could add a `NaN` here too, to be able to split interiors? 
The array of Points won't be of much use if we can't differentiate the geometry element(s) it is composed of.

visr: I'm not sure what is best for AbstractPlotting, but I think `NaN` here would be a bad idea. Decompose here simply means "give me all the points that make up this polygon". This is consistent with the other decompose methods, and this consistency is key here. If you want to be able to differentiate the exterior and interiors, it is easy to call decompose on those parts individually.

Sov-trotter: Oh. I had missed out how other decompose methods behave. Then it's better to return all the all the points.

Sov-trotter:
@visr @SimonDanisch  can you take a look?

src/interfaces.jl
```diff
@@ -112,6 +112,20 @@ function decompose(::Type{T}, primitive) where {T}
...
 
+function decompose(::Type{P}, pol::Polygon) where {P<:AbstractPoint}
+    if isempty(pol.interiors)
+        return decompose(P, pol.exterior)
+    else
+        arr = P[]
+        push!(arr, decompose(P, pol.exterior)...)
+        for i in pol.interiors
+            push!(arr, decompose(P, i)...)
+        end
```
visr, line 123: Splatting such potentially large arrays can have a negative performance impact. It's best to use `append!` instead.
And you can just start with the exterior like above, `arr = decompose(P, pol.exterior)`

Sov-trotter: I actually did that, `arr = decompose(P, pol.exterior)` but seems like REPL's new soft scope might be causing problems here. If I call decompose on poly repeatedly, the interiors are appended to the previous result everytime instead of creating a new array. 
Screenshot from 2020-08-08 18-39-26

Sov-trotter:
```julia
using GeometryBasics

poly = Polygon(LineString(Point{2, Int}[(5, 1), (3, 3), (4, 8), (1, 2), (5, 1)]), [LineString(Point{2, Int}[(2, 2), (3, 8),(5, 6) , (3, 4), (2, 2)]), LineString(Point{2, Int}[(3, 2), (4, 5),(6, 1) , (1, 4), (3, 2)])])

decompose(Point, poly)
15-element Array{Point{2,Int64},1}:

decompose(Point, poly)
25-element Array{Point{2,Int64},1}:

decompose(Point, poly)
35-element Array{Point{2,Int64},1}:

```
and so on..

visr: Oh yeah, good point, hadn't thought of that. It's not a scope issue, it's just that `arr = decompose(P, pol.exterior)` doesn't return a copy of the polygon points, but the original ones. So adding to that is a bad idea. Could also be solved by `arr = copy(decompose(P, pol.exterior))`.

Sov-trotter: So this is weird but I tried to get to the root of the problem and something seems really off. 
So here's what's happening:
`arr = decompose(P, pol.exterior)` is actually creating some kind of pointer to `pol.exterior`. Because of which the exteriors are growing in size(the interiors are added into exteriors.)

```julia
pol = Polygon(LineString(Point{2, Int}[(5, 1), (3, 3), (4, 8), (1, 2), (5, 1)]), [LineString(Point{2, Int}[(2, 2), (3, 8),(5, 6) , (3, 4), (2, 2)]), LineString(Point{2, Int}[(3, 2), (4, 5),(6, 1) , (1, 4), (3, 2)])])

pol.exterior
4-element LineString{...

decompose(Point{2, Int}, pol)
15-element Array{Point{2,Int64},1}:

pol.exterior
14-element LineString{...

decompose(Point{2, Int}, pol)
25-element Array{Point{2,Int64},1}:

pol.exterior
24-element LineString{....
```

Now I don't know why is it appending to the exterior here(it's a fast way to modify existing `linestring` though haha).

Sov-trotter: So every time I call decompose 
```
for i in pol.interiors
    append!(arr, decompose(P, i))
end
```

appends it to exteriors. :/

Sov-trotter: But like you said `copy()` works.

visr: Yeah indeed, that is what I meant with "doesn't return a copy of the polygon points, but the original ones". It's like this:

```julia
julia> a = [1]
1-element Array{Int64,1}:
 1

julia> b = a
1-element Array{Int64,1}:
 1

julia> push!(b, 2)
2-element Array{Int64,1}:
 1
 2

julia> a
2-element Array{Int64,1}:
 1
 2
```

Sov-trotter:
Really sorry for pestering you this way, but I'd be nice if you can take a look at this PR and #79?

Issue #50, opened 2020-05-22
heterogeneous geometry collection


visr:
Sorry for spamming the issues, but just want to get some of them out of my head :)

This issue is somewhat similar to #49, but that is about features (geometries + properties) with heterogeneous geometry types in a feature collection.

Some formats, like GeoJSON for instance, also support heterogeneous geometry collections, which are multigeometries like the multipoint or multipolygon we already have, but can consist of different geometry types. In GeoJSON it is called [GeometryCollection](https://tools.ietf.org/html/rfc7946#section-3.1.8). I'm wondering how we could represent them. It may be nice to have a geometry type like this in or based on this package, even if it will be hard to optimize.

EDIT: for now I just went with a bare Vector, though it may be nice to have that vector in a GeometryCollection type: 
https://github.com/visr/GeoJSONTables.jl/blob/4104e66a638814d77ef98af1d205450549361519/test/basics.jl#L74-L76

asinghvi17:
I guess that this isn't really needed with GeoInterface geometry collection wrappers now?

Issue #71, opened 2020-07-21
Redesigning Metadata processing and creating of meta-geometries 


Sov-trotter:
The current metadata handling features have been really interesting and there has been interest to expand it's functionality #50 #49. 
The above mentioned issues have been made keeping in mind the heterogeneous nature of geometry/properties data. The present methods don't support this fully. 
If I am correct, the way of creating meta-geometries is putting geometries and properties together into a StructArray, which works well for single geometries/properties or homogeneous geometries. The problem arises when we need a heterogeneous geometry with heterogeneous type like #49. 

Now this functionality has been tested and implemented [here](https://github.com/visr/GeoJSONTables.jl/pull/3) we wish to extend it to GeometryBasics so that other use cases too can benefit from this API.

Sov-trotter:
So the above GeoJSONTables solution works well for a list of Features(meta-geometries). Currently, what all can we have?
* A Feature(meta-geometry struct that has properties and geometries together)
* A StructArray that holds the metageometry
I am not sure if we want to implement geometry specific metageometry constructors(we might as well).

sjkelly:
I think this is quite needed for a lot of geometries and algorithmic shortcuts. Is the current proposal non-breaking?

Sov-trotter:
It is breaking indeed. We are moving from something like `PointMeta` to `Meta{Point}`. This is to solve #49.

ffreyer:
Closing this since we removed the meta interface

Issue #49, opened 2020-05-22
heterogeneous feature collection


visr:
I understand that the way to have geometries with attributes behave like tables in this package is to create a StructArray for the collection.

Do you have any idea what we could do when we want to represent different geometry types? Most commonly a dataset is all of the same type, but there are exceptions, and it would be nice to be able to represent them as well.

This understandably fails:
```julia
point = meta(Point(3, 1), city="Abuja", rainfall=1221.2)
polygon = meta(Polygon(Point{2, Int}[(3, 1), (4, 4), (2, 4), (1, 2), (3, 1)]), city="Borongan", rainfall=4114.8)

StructArray([point, polygon])  # ArgumentError: type does not have a definite number of fields
```

If I try to throw it into for instance a TypedTables.Table, it works fine, accepting geometry as a Vector{Any}.

```julia
using TypedTables
Table(
    geometry=[Point(3, 1), Polygon(Point{2, Int}[(3, 1), (4, 4), (2, 4), (1, 2), (3, 1)])],
    city=["Abuja", "Borongan"],
    rainfall=[1221.2, 4114.0],
)
```

I know this is more of a basic StructArrays vs TypedTables question, and understand that with a Vector{Any} things will be slower. But it would be nice to have a "GeometryBasics" table solution for this as well.

EDIT: concrete example here: https://github.com/visr/GeoJSONTables.jl/blob/4104e66a638814d77ef98af1d205450549361519/test/basics.jl#L97-L101

piever:
Came here from https://github.com/JuliaArrays/StructArrays.jl/issues/135.
I'm slowly wrapping my head around this use case. I am slowly realizing that what I said there may be a bit misleading, as you _do not_ want to do the "struct of array, array of structs" transform on the geometries, but just store an array of custom structs (possibly heterogeneous).

In that case, you can use `StructArray` just like any other table, so the following would work:

```julia
StructArray(
    geometry=[Point(3, 1), Polygon(Point{2, Int}[(3, 1), (4, 4), (2, 4), (1, 2), (3, 1)])],
    city=["Abuja", "Borongan"],
    rainfall=[1221.2, 4114.0],
)
```

[...]

PR #74, opened 2020-07-28
Support heterogeneous features

Sov-trotter:
I haven't removed the old meta implementation as planned due to some issues with `MeshMeta`. But for the new implemetation, tests are almost written. 
We just need to fix the Mesh issue, add some other helpful constructors.


Sov-trotter:
#71

visr:
@SimonDanisch the basic reason for going down this path is https://github.com/JuliaGeometry/GeometryBasics.jl/issues/49#issuecomment-651203611. This PR is only a draft, but do you in general agree with the proposal of moving from `PointMeta` to `Meta{Point}`? Main downside is that `Meta{Point}` can no longer subtype `AbstractPoint`.

We are going with the name `Feature` here which is the GeoJSON concept it corresponds to, and to avoid the clash with `Base.Meta`, but the name is of course open for discussion. A `Feature` can be seen as a row in a table. In GeoJSON this table is a `FeatureCollection`. We could go with `StructArray{Feature}` here or wrap that in a `FeatureCollection` struct.

SimonDanisch:
>Main downside is that Meta{Point} can no longer subtype AbstractPoint.

Oh, yeah...that was 100% the reason, that I didn't go with it :D 

I think having `MetaPoint` be an `AbstractPoint` is pretty important ;) 
So one way of keeping that would be, to have `Meta{X}` a fallback, for types, that weren't explicitly created by the macro?

visr:
I think that would work for us :)

Sov-trotter:
closes #50 #49

Sov-trotter:
I think this is ready to be merged?
Are there anymore methods that I should add?

SimonDanisch:
@visr, can you give this a review as well? :)

visr:
Yes, I'll have a look in a day or two.

src/metadata.jl
```diff
@@ -190,3 +190,70 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
 @meta_type(Mesh, mesh, AbstractMesh, Element <: Polytope)
 Base.getindex(x::MeshMeta, idx::Int) = getindex(metafree(x), idx)
 Base.size(x::MeshMeta) = size(metafree(x))
+
+
+"""
+`Feature` type acts same as Meta method 
+The difference lies in the fact that it is designed to handle
+heterogeneous types.
```
visr, line 198: Perhaps good to explain this a bit more, since a geometry cannot be heterogenous. You could list the advantage: you can put any type in there, and the disadvantage: it is no longer a subtype in the way `MetaPoint` and others are.

Would also be nice if you could add the signature of the different constructors on top, and an example below, like here: https://docs.julialang.org/en/latest/manual/documentation/#Documentation-1

src/metadata.jl
```diff
@@ -190,3 +190,70 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+end
+
+Feature(x; kwargs...) = Feature(x, values(kwargs))
+
+"""
+Frees the Feature out of metadata
+i.e. returns and array of geometries
+"""
+function metafree(F::Feature)
+    getproperty(F, :main)
```
visr, line 215: Best to use `getfield` here, like you do below, to make it robust to `getproperty` overloading. And a small detail, but since variable names are lowercase, capital `F`, could be confusing here to some people. Could be just `x`.

src/metadata.jl
```diff
@@ -190,3 +190,70 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+end
+metafree(x::AbstractVector{<: Feature}) = [getproperty(i, :main) for i in x]
```
visr, line 217: Could be `= map(metafree, x)`, same applies for `meta` below.

src/metadata.jl
```diff
@@ -190,3 +190,70 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+"""
+Returns the metadata of a Feature
+"""
+function meta(x::Feature)
+    getfield(x, :meta)
+end
+meta(x::AbstractVector{<: Feature}) = [getproperty(i, :meta) for i in x]
+
+# helper methods
+Base.getproperty(f::Feature, s::Symbol) = s == :main ? getfield(f, 1) : s == :meta ? getfield(f, 2) : getproperty(getfield(f, 2), s)
```
visr, line 228: I personally find these a bit hard to read. Could you use the multiline `function f(x) ... end` syntax and if statements to split it up a bit? The `getfield(f, 1)` is clever, didn't know that you could get the first field that way, but I'd just use the `meta` and `metafree` functions that you already defined. You can also copy the existing `getproperty` overload in the macro.

src/metadata.jl
```diff
@@ -190,3 +190,70 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+end
+
+# generate an instance of Feature type
+function StructArrays.createinstance(::Type{F}, x, args...) where {F<:Feature}  
+     T , names, types = getnamestypes(F)
+     Feature(x, NamedTuple{names, types}(args))
+end
+
+"""
+Accepts an Array/iterator of Features and put it into a StructArray
```
visr, line 245: "Array/iterator" could be "iterable"

src/metadata.jl
Sov-trotter: Ah yeah. I had saved it for the docs but will add on here also.

Sov-trotter: >Would also be nice if you could add the signature of the different constructors on top. 

Sorry, Could you point out the constructors you're referring to here

Sov-trotter:
I think GeometryMeta is good to have?

src/GeometryBasics.jl
```diff
@@ -35,7 +35,8 @@ module GeometryBasics
     export GLTriangleFace, GLNormalMesh3D, GLPlainTriangleMesh, GLUVMesh3D, GLUVNormalMesh3D
     export AbstractMesh, Mesh, TriangleMesh
     export GLNormalMesh2D, PlainTriangleMesh
-
+    export MetaT, collect_MetaT
```
visr, line 38: Since function names should be lowercase by convention, it's best to rename this. Using `collect` in the name may cause people to think it works similar to `Base.collect`, which is not the case. Something like `meta_table` perhaps?

src/metadata.jl
```diff
@@ -190,3 +192,98 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
 @meta_type(Mesh, mesh, AbstractMesh, Element <: Polytope)
 Base.getindex(x::MeshMeta, idx::Int) = getindex(metafree(x), idx)
 Base.size(x::MeshMeta) = size(metafree(x))
+
+
+"""
+
+    MetaT(::type{T}, Names, Types)
```
visr, line 199: This signature should represent how users can construct this type. These are the two ways:
```
MetaT(geometry, meta::NamedTuple)
MetaT(geometry; meta...)
```

src/metadata.jl
```diff
@@ -190,3 +192,98 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+    
+Returns a `MetaT` that holds a geometry and it's metadata
```
visr, line 201: typo: its

src/metadata.jl
```diff
@@ -190,3 +192,98 @@ Base.size(x::MultiPolygonMeta) = size(metafree(x))
...
+"""
+
+    metafree(x::MetaT)
+    metafree(x::Array{MetaT})
+
+Free the MetaT from metadata
+i.e. returns the geometry/array of geometries
+"""
+function metafree(x::MetaT)
+    getfield(x, 1)
```
visr, line 232: Sorry to be so nitpicky, but I'd prefer `getfield(x, :main)` here, and `getfield(x, :meta)` below. It's easier to recognize the name rather than the position of the field as a reader, and will not be forgotten if we decide to rename fields later on.

src/metadata.jl
Sov-trotter: Yeah I was actually confused here, Thanks! :)

Sov-trotter:
I am extremely sorry for the scattered PR's(both here and in AbstractPlotting). But here's a sequence in which you might want to review/merge them:

1. Firstly this one
2. #73 
3. https://github.com/JuliaPlots/AbstractPlotting.jl/pull/479 (Depends on the `decompose()` method in #73 )
4. https://github.com/JuliaPlots/AbstractPlotting.jl/pull/486 (Depends on the `convert_arguments` method for `LineString` in the above PR)

Sov-trotter:
Hey @SimonDanisch if the changes are fine then this can be merged?

SimonDanisch:
Sorry for being a bit slow! Thanks for you contribution!

PR #78, opened 2020-08-12
Add min Euclidean distance from rectangle

ambyld:
This function is already exported, but the definition was missing.


SimonDanisch:
Thanks! Must have gotten lost in some refactor, I suppose

PR #87, opened 2020-09-22
Remove Quad{T} in favor of Rect3D{T}

juliohm:
This PR removes `Quad{T}` in favor of the new `Rect3D{T}`.


SimonDanisch:
I think we can remove this...seems like it isn't actually used anywhere

PR #93, opened 2020-09-25
[WIP] Cleanup Vec and Point types

### LINKED ISSUE #91: Vec, Point, SVector, ... in the codebase
As I continue to study the source code to suggest improvements, I would like to clean up all the variations around the notion of a point as a static n-dimensional vector. Currently, the code uses `Vec`, `Point`, `SVector` in different places, and it is hard to understand if there is an actual reason behind these variations or if it is just a result of the code evolving quickly. Can we pick a single name and stick to it? I can submit a PR with a single name if that is ok.

juliohm:
Fix #91 

1. First we convert all occurrences of tuples, SVector, Vec, ... into a single `const Vec = SVector` to be used consistently across the codebase. Time for us to sign a contract with blood :drop_of_blood: to never introduce these variations again :) In this simple exercise I already fixed a conceptual bug where a vector was being used in the place of a point in `Base.in(p::Point, r::Rect)`. Also, bye bye `VecTypes`, we don't need you anymore.

2. Second we remove the `Particle` type as it is not really a geometry primitive. Given that the concept of a particle resembles points and vectors, we can get rid of it here as well.


test/runtests.jl
```diff
@@ -464,8 +464,8 @@ end
     @test iterate(texturecoordinates(r2)) == ((0.0, 1.0), ((0.0, 2), (1.0, 2)))
     r3 = Rect3D(0.0, 0.0, 1.0, 1.0, 2.0, 2.0)
     @test iterate(texturecoordinates(r3)) == ([0, 0, 0], 2)
-    uv = decompose_uv(m)
-    @test boundbox(Point.(uv)) == Rect(0, 0, 1, 1)
+    # uv = decompose_uv(m)
+    # @test boundbox(Point.(uv)) == Rect(0, 0, 1, 1)
```
juliohm, line 468: @SimonDanisch could you please help with this test? It is producing a stack overflow and I couldn't figure out the logic.

test/geometrytypes.jl
```diff
@@ -85,8 +85,8 @@ end
         m = normal_mesh(s)# just test that it works without explicit resolution parameter
         @test m isa GLNormalMesh
 
-        muv = uv_mesh(s)
-        @test boundbox(Point.(texturecoordinates(muv))) == FRect2D(Vec2f0(0), Vec2f0(1.0))
+        # muv = uv_mesh(s)
+        # @test boundbox(Point.(texturecoordinates(muv))) == FRect2D(Vec2f0(0), Vec2f0(1.0))
```
juliohm, line 89: @SimonDanisch could you please help with this test? It is producing a stack overflow and I couldn't figure out the logic.

src/primitives/rectangles.jl
```diff
@@ -260,8 +260,8 @@ function Base.:(*)(m::Mat{4,4,T}, h::Rect{3,T}) where {T}
     return Rect{3,T}(_vmin, _vmax - _vmin)
 end
 
-Base.:(-)(h::Rect{N,T}, move::Number) where {N,T} = h - Vec{N,T}(move)
-Base.:(+)(h::Rect{N,T}, move::Number) where {N,T} = h + Vec{N,T}(move)
+Base.:(-)(h::Rect{N,T}, move::Number) where {N,T} = h - Vec{N,T}(ntuple(i->move,N))
```
SimonDanisch, line 263: just use `.-` ?

SimonDanisch: Oh I see what happened here... the constructor is gone... Oh no. This is my favorite constructor :D

juliohm: I tried adding the dot in front of minus, but the stack overflow still happened. More tests fail when I add the dot. Any other possible fix?

Regarding the constructor with a single value, the plan is to add a helper function later like `nval(n, v) = Vec(ntuple(i->v,n))` so one could do `nval(3, 1.0f0)`. This is a similar pattern we've fixed in that `boundbox` PR. I think the constructors are being abused :D We just need a new function name for this operation. So far the better name I could come up with is `nval`, but please let me know if you have a better name in mind. We could also rename the `unit` helper to `basis` to avoid confusion with `nval(n, 1)`.

test/geometrytypes.jl
juliohm: @SimonDanisch did you have a chance to take a look into this? I can try debug it further today, but it will certainly take much longer to discover what is happening given that I am not familiar with all the parts of the codebase yet.

After these tests pass, I will proceed to the last (and most challenging) step of this PR, which is the point type.

SimonDanisch: To be honest, I'm not entirely sure anymore, that we should do this - I always forget how breaking the constructor changes are, and how much I use them...
Just imagine this little refactor with you debugging experience you have right now, were you're giving up and referring that to me, and take that by ~20x - that'd approximately be the work that will be put on me if we go through with this.

So maybe we should just clean up the implementation a bit, and make sure, that Point & Vec are used consistently!

juliohm:
Simon I really think that this would be the best thing to do in the long
run. There are so many hidden bugs and the maintenance burden is all on you
because few people can digest the codebase with so many constructors doing
much more than just constructing. I'm proposing to become a maintainer if
you didn't notice:) it will be much harder for me and others to maintain
with these changes not merged. Imagine when we start to apply
transformations to these geometries, these constructors may break in non
Euclidean manifolds. It is just too hard to keep up with the details and
variations. Also as I mentioned we can have nval and other utility
functions for the most common use cases? I'm aiming big here with non
Euclidean geometries,  doing calculations on the sphere directly. These
goals won't be achieved with the current state of affairs.

SimonDanisch:
I'm sorry but I'm just realistic - and the reality is, that i'm super bottle necked. 
Before burdening me with a huge amount of work, I'll need much more evidence than:
"""
Vec3f0(1) == Vec3f0(1, 1,1) is a huge maintenance burden and there are so many hidden bugs with it
"""
I can totally see, that GeometryBasics has a lot of hidden bugs - since it's not tested very well^^
But blaming this on the few methods defined on Vec on top of the methods on SVector seems a bit drastic!

juliohm:
I meant that the use of vectors and points is being interchanged
(incorrectly) and that tuples and vectors have constructors that broadcast
entries (incorrectly like above) which causes more bugs. Cleaning these use
cases is a first step towards a more robust base package. These convenience
constructors should really be convenience functions like we did for
boundbox. We were originally abusing the Rect constructor and it took me
some time to follow the dispatch of the bounding box use cases. So you are
imposing a maintaince burden on we all together when these constructors are
used everywhere. Makes sense? It is much easier to track down boundbox and
nval in the codebase than search for all call sites of Vec with all equal
coordinates.

SimonDanisch:
I agree with the general consent... And don't get me wrong, I'm surely happy about more maintainers & usage.
But it would be much easier on me, to do this step by step. So, e.g. introduce a `nval` function, while keeping the constructors around.

>search for all call sites of Vec with all equal coordinates.

Btw, I'm not sure that's very convincing :D The only refactor that comes to mind where you'd search for that is the one you're currently doing - which I want to avoid... Or are there other refactors, where one would want to replace such constructors?

Btw, I overlooked, that you only changed the filename from `boundbox->boundingboxes` . Of course, the desire to have unshortened names includes function names, and not only files ;)

juliohm:
> I agree with the general consent... And don't get me wrong, I'm surely happy about more maintainers & usage.
> But it would be much easier on me, to do this step by step. So, e.g. introduce a `nval` function, while keeping the constructors around.

Sorry, I think I wasn't clear about the refactoring plan. The idea is to clean first, make sure every test still pass, and only then introduce `@depwarn` deprecation warnings correcting for the new behavior. This seems to solve all your concerns about maintenance? It was never part of the plan to create more work on your side.

Let me reiterate again the importance of cleaning this up as soon as possible. Take this PR as an example. It is an evidence that the abuse of the constructor with a single value and possible broadcasting behavior is more harmful than helpful. The only thing  I did in this PR was: "let's clean all variations of vectors to a single SVector to be used consistently across the codebase". Now this simple change is causing tests to stack overflow. Why? Very likely the constructors with single values are being used in a place where SVectors are now being passed and somehow things explode. Another evidence that the cleanup is important is finding code like the following in the codebase:

```julia
function widths(x::AbstractRange)
    mini, maxi = Float32.(extrema(x))
    return maxi - mini
end
```

First it assumes that we support `AbstractRange` to store widths. Second it explicitly converts to Float32 because it is probably only being used in a Makie.jl context with single precision for plotting. So the `Vecf0` constructor is being used too much downstream just because it exists. Ideally we would clean these occurrences as well. So to answer your second question:

>Or are there other refactors, where one would want to replace such constructors?

Yes, ideally we would clean up all this variations around the concept of a static vector for storing coordinates. The result of not doing so early is that by default we will have more headache fixing bugs downstream. Again, the plan is to do it gradually with deprecation warnings.

>Btw, I overlooked, that you only changed the filename from boundbox->boundingboxes . Of course, the desire to have unshortened names includes function names, and not only files ;)

You mean that we should also rename `boundbox` to `boundingbox`? I can work on it after this PR here is ready to be merged. For that I would really appreciate your help with the tests above that are failing. As I become more familiar with the concepts in the codebase I will be able to fix more things by myself, but right now there is a lot to work out before going down the path of texturecoordinates and meshes, which is what is currently causing the stack overflow.

juliohm:
Came back to this PR today. I am trying to fix the 2 tests that failed. I am trying to digest the logic of this function:

```julia
function decompose(UVT::Union{UV{T},UVW{T}}, primitive) where {T}
    # This is the fallback for texture coordinates if a primitive doesn't overload them
    # We just take the positions and normalize them
    uv = texturecoordinates(primitive)
    if uv === nothing
        # If the primitive doesn't even have coordinates, we're out of options and return
        # nothing, indicating that texturecoordinates aren't implemented
        positions = decompose(Point, primitive)
        positions === nothing && return nothing
        # Let this overlord do the work
        return decompose(UVT, positions)
    end
    return collect_with_eltype(T, uv)
end
```

You are trying to compute the UV or UVW coordinates of the vertices in the mesh. However, it seems that the return of `texturecoordinates` is nothing for the `m = mesh(Sphere(Point3f0(0), 1))`, and we enter in an infinite recursion where `decompose` calls itself. Can you please clarify the logic here?

juliohm:
Ok, all tests are passing now after I fixed another BUG in the source code related to another misuse of point vs. vector. I hope this is enough evidence that we really need to clean things up.

@SimonDanisch could you please share your point of view about these changes after our conversations? Do you plan to merge them into master at some point?

juliohm:
Should I fork the project and proceed on my own?

SimonDanisch:
Sorry, I'm a bit on a break right now! 

>Do you plan to merge them into master at some point?

As I said, I'll be fine with most refactors, that don't mean a lot of work for me!

juliohm:
> As I said, I'll be fine with most refactors, that don't mean a lot of work for me!

Does that mean that all the changes in this PR are acceptable? I just want to double check that you have plans to merge these at some point before I continue with the refactoring of the code and additional fixes.

juliohm:
I will merge these changes to the `cleaned` branch that I am working on, and will continue submitting PRs for it in the next days. When you are back from the break, you let me know if the branch is ok to be merged back into master.

PR #115, opened 2020-12-26
Add isempty

timholy:
This will be used to avoid boxing (in the sense of `Core.Box`, a.k.a. julia#15276) in https://github.com/JuliaPlots/AbstractPlotting.jl/blob/6346fe87f18bb1df09f0065f0ad6b2b699c20b94/src/layouting/boundingbox.jl#L170-L199. I took the liberty of bumping the version to encourage a new release.


SimonDanisch:
Thanks :)

PR #122, opened 2021-01-21
Fix mesh(::AbstractPolygon) in case faces returns nothing

knuesel:
When `faces` returns `nothing`, `mesh(::Meshable)` falls back on `decompose(facetype, positions)`. The fallback is missing from `mesh(::AbstractPolygon)` so the following fails:

```
julia> t = Triangle(Point2f0(0), Point2f0(1), Point2f0(1,0))
Triangle(Float32[0.0, 0.0], Float32[1.0, 1.0], Float32[1.0, 0.0])

julia> GeometryBasics.mesh(t)
ERROR: MethodError: no method matching Mesh(::Array{Point{2,Float32},1}, ::Nothing)
```

This PR adds the fallback to the `AbstractPolygon` case.


SimonDanisch:
Thanks

Issue #123, opened 2021-01-21
Invalid keyword argument syntax in latest version


kylejbrown17:
The latest version (v0.3.6) has a few keyword syntax errors that cause a LoadError when I try to load the package.

```julia
julia> ]

(@v1.4) pkg> activate LDrawParser
 Activating environment at `~/.julia/dev/LDrawParser/Project.toml`

(LDrawParser) pkg> status GeometryBasics
Project LDrawParser v0.1.0
Status `~/.julia/dev/LDrawParser/Project.toml`
  [5c1252a2] GeometryBasics v0.3.6

julia> using GeometryBasics
[ Info: Precompiling GeometryBasics [5c1252a2-5f33-56bf-86c9-59e7332b4326]
ERROR: LoadError: LoadError: syntax: invalid keyword argument syntax "pointtype"
Stacktrace:
 [1] top-level scope at /home/kylebrown/.julia/packages/GeometryBasics/bMGVD/src/meshes.jl:100
 [2] include(::Module, ::String) at ./Base.jl:377
 [3] include(::String) at /home/kylebrown/.julia/packages/GeometryBasics/bMGVD/src/GeometryBasics.jl:1
 [4] top-level scope at /home/kylebrown/.julia/packages/GeometryBasics/bMGVD/src/GeometryBasics.jl:22
 [5] include(::Module, ::String) at ./Base.jl:377
 [6] top-level scope at none:2
 [7] eval at ./boot.jl:331 [inlined]
 [8] eval(::Expr) at ./client.jl:449
 [9] top-level scope at ./none:3
in expression starting at /home/kylebrown/.julia/packages/GeometryBasics/bMGVD/src/meshes.jl:100
in expression starting at /home/kylebrown/.julia/packages/GeometryBasics/bMGVD/src/GeometryBasics.jl:22
ERROR: Failed to precompile GeometryBasics [5c1252a2-5f33-56bf-86c9-59e7332b4326] to /home/kylebrown/.julia/compiled/v1.4/GeometryBasics/lB452_nEEYK.ji.
Stacktrace:
 [1] error(::String) at ./error.jl:33
 [2] compilecache(::Base.PkgId, ::String) at ./loading.jl:1272
 [3] _require(::Base.PkgId) at ./loading.jl:1029
 [4] require(::Base.PkgId) at ./loading.jl:927
 [5] require(::Module, ::Symbol) at ./loading.jl:922
[...]

PR #124, opened 2021-01-21
Fixed the invalid keyword syntax that was causing a LoadError. 

kylejbrown17:
All tests passing (Julia 1.4, Ubuntu 16.04).


kylejbrown17:
Forgot to link to the issue #123

visr:
Ah, thanks, good to keep 1.4 compat, since we specify compatibility with julia 1.3+.

Issue #121, opened 2021-01-21
Area of 2D triangle can be negative


knuesel:
For example:

```
julia> area(Point2f0[[0,0], [0,1], [1,1]])
-0.5f0
```

The relevant `area` method is here:
https://github.com/JuliaGeometry/GeometryBasics.jl/blob/24d0ee53b0d2ff3eca5bc3a8c864ad53d534df76/src/triangulation.jl#L36

I think there are two issues with this method:

1. It's the generic implementation for N dimensions, but it only works for N=2, as it assumes that `cross` exists and returns a scalar.
2. It's missing an `abs` call on the result.

Regarding the first point: there's already a specific method for N=3, and we can make one for N=2. Do we need a generic method for N>3?

SimonDanisch:
Good point!

> Do we need a generic method for N>3?

Probably not :D 
Sounds like a good case for only specializing on 2 & 3 .

knuesel:
Looks like the negative area was intentional:

https://github.com/JuliaGeometry/GeometryBasics.jl/blob/24d0ee53b0d2ff3eca5bc3a8c864ad53d534df76/src/triangulation.jl#L127

Should we leave it like this (and document it)?

SimonDanisch:
Oh, good catch :D Yeah, I guess the area function is mostly used inside GeometryBasics to figure out the winding order.

knuesel:
I made the PR for the doc change, but it is a bit inconsistent to return a signed area for 2D an a positive area for 3D... Instead, we could have `area` return always a positive value, and a new `oriented_area` that returns a signed value for 2D and a vector for 3D?

PR #125, opened 2021-01-22
area and MetaFree fixes

knuesel:
Some fixes for `area` following discussion in #121

* `area(::AbstractVector{AbstractPoint}})` is now documented
* In particular, the doc says that a signed value is returned in the 2D case
* Fixes for custom indices (OffsetArrays is added as a test dependency)
* The 3D case now accepts vectors of any `AbstractPoint{3}` type (the 2D case already did)

There's also a broken `area` test for a vector of `PointMeta{3}`. I couldn't make it work because it needs `zero(typeof(PointMeta(0,0,0; a=1))` and it's not clear how to implement that: How should we make the zero value of the metadata?

The PR includes a fix for a related bug: `MetaFree(typeof(PointMeta(0,0,0; a=1))` used to return `3`. Now it returns `Point{3,Int64}`, I think that's what was intended? (the fix isn't enough to make `area` work but I included it in the PR since it took me a while to figure out).

I also added Julia 1.3 in the CI test matrix, to avoid a repeat of #123 😊

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
+          - '1.3'
           - '1'
           - 'nightly'
         os:
--- a/Project.toml
+++ b/Project.toml
@@ -19,8 +19,9 @@ Tables = "0.2, 1"
 julia = "1.3"
 
 [extras]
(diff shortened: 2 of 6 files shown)


SimonDanisch:
Thank you! :)

PR #131, opened 2021-02-07
Export `HyperRectangle`

ali-ramadhan:
Not sure why it's not being exported but I found it useful on its own so thought it might be good to export.


SimonDanisch:
Why not ;)

PR #137, opened 2021-03-04
Warn, don't error, on precompile failures

timholy:
I don't know of any problems currently, but hopefully this future-proofs you against problems.


SimonDanisch:
Thank you ❤️ !

PR #138, opened 2021-03-08
fuller StructArrays integration

piever:
In StructArrays 0.5, the interface has been slightly modified (see [NEWS entry](https://github.com/JuliaArrays/StructArrays.jl/blob/master/NEWS.md#version-050)), so now it is necessary to overload `component` as well for stucts with custom layout.

This allows better StructArrays integration and makes things like `setindex!` work out of the box with `StructArray`s of meta geometries.

I've also added tests to make sure the full interface is implemented.


SimonDanisch:
Nice! :) Thanks a lot!

PR #146, opened 2021-06-28
update structarrays compat

piever:
In StructArrays 0.6, StaticArrays are treated the same way as the underlying `Tuple` in terms of data storage, so one should pass the tuple of columns directly to the constructor, without having to nest `StructArray`s.

I'm dropping compatibility with the older version, because it seems tricky to support both.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,8 @@ jobs:
   docs:
     name: Documentation
     runs-on: ubuntu-latest
+    env:
+      JULIA_PKG_SERVER: ""
     steps:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
@@ -52,7 +54,6 @@ jobs:
       - run: |
           julia --project=docs -e '
             using Pkg
-            Pkg.update()
             Pkg.develop(PackageSpec(path=pwd()))
             Pkg.instantiate()'
       - run: |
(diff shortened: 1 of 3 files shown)


piever:
Tests pass locally, but I'm not sure why the documentation build does not find version StructArrays 0.6

SimonDanisch:
Hmmm weird indeed.. Can you try to add an `Pkg.up()` to the docs ci stage?

SimonDanisch:
image

Those other CI stages should run almost exactly the same !? :D

piever:
Yeah, not too sure how to debug this... Maybe it's just a matter of waiting a bit and trying again :)

I thought maybe it doesn't find the version because StructArrays 0.6 requires julia 1.3 minimum, but the docs seem to run on 1.6, so that shouldn't be it

knuesel:
This is an issue with the PkgServer protocol serving an outdated registry. At first I couldn't reproduce this locally because I had the old Git-clone version of the registry in `~/.julia/registries/General`. After removing it and installing StructArrays I got version 0.5.1.

There's a discussion at https://github.com/JuliaRegistries/General/issues/16777 and workarounds have been implemented e.g. at https://github.com/julia-actions/julia-runtest/pull/17 and https://github.com/invenia/PkgTemplates.jl/pull/296.

For example in our case replacing the `Pkg.update()` line in `ci.yml` with `ENV["JULIA_PKG_SERVER"] = ""` would do the trick.

PR #148, opened 2021-07-26
Added functions to get distances to primatives and meshes

weymouth:
Created new file distances.jl which computes the absolute_distance and 
the signed_distance to triangle meshes, HyperSpheres and HyperRects. 
These functions have been exported.

The sphere and rect signed functions are straightforward, and there is a 
fallback for absolute_distance to equal abs(signed_distance).

Dealing with triangles and triangle meshes is less straightforward. The 
absolute distance to a triangle is a well defined (if complex) function, 
but the signed distance isn't. I've used the signed distance to the 
plane aligned with the triangle. The signed distance to a mesh of 
triangles is then the signed distance to the convex union of those 
planes. This is simple, but obviously has drawbacks.

Alternatively, the absolute distance to a mesh of triangles is robust, 
and could be supplimented with a global operation (like the winding 
number) to determine a sign. 

I've added a 27 tests to runtests which illustrate some of these issues.

### DIFF
--- a/src/distances.jl
+++ b/src/distances.jl
@@ -1,0 +1,117 @@
+#=
+Functions to compute distances from points to primatives
+=#
+
+using LinearAlgebra: normalize,norm,⋅
+"""
+    closest_point_to_tri(p, a, b, c)
+
+This method from Ericson, "Real-time collision detection"
+2005, doesn't require the normal or cross products.
+"""
+@fastmath function closest_point_to_tri(p,a,b,c)
+    # is point `a` closest?
+    ab,ac,ap = b-a,c-a,p-a
+    d1,d2 = ab ⋅ ap, ac ⋅ ap
+    d1 <= 0 && d2 <= 0 && return a
+
+    # is point `b` closest?
+    bp = p-b
+    d3,d4 = ab ⋅ bp, ac ⋅ bp
+    d3 >= 0 && d4 <= d3 && return b
+
+    # is point `c` closest?
+    cp = p-c
+    d5,d6 = ab ⋅ cp, ac ⋅ cp
+    d6 >= 0 && d5 <= d6 && return c
+
+    # is segment `ab` closest?
+    vc = d1*d4 - d3*d2
+    if (vc <= 0 && d1 >= 0 && d3 <= 0)
+        d1==d3 && @show d1,d3
+        return a + ab * d1 * inv(d1 - d3)
+    end
+
+    # is segment `ac` closest?
+    vb = d5*d2 - d1*d6
+    if (vb <= 0 && d2 >= 0 && d6 <= 0)
+        return a + ac * d2 * inv(d2 - d6)
+    end
+
+    # is segment `bc` closest?
+    va = d3*d6 - d5*d4
+    if (va <= 0 && d4 >= d3 && d5 >= d6)
+        return b + (c - b) * (d4 - d3) * inv(d4 - d3 + d5 - d6)
+    end
+
+    # closest is interior to `abc`
+    denom = inv(va + vb + vc)
+    v,w = vb * denom, vc * denom
+    return a + ab * v + ac * w
+end
+"""
+    closest(p,tri::Triangle)
+
+Determine the closest point on triangle `tri` to point `p`.
+"""
+closest(p,tri::Triangle) = closest_point_to_tri(p,tri.points.data...)
+
+"""
+    absolute_distance(p,tri::Triangle)
+
+Determine the absolute distance from point `p` to the closest point on triangle `tri`.
+"""
+absolute_distance(p,tri::Triangle) = norm(p-closest(p,tri))
+
+"""
+    signed_distance(p,tri::Triangle)
+
+Determine the signed distance from point `p` to the plane defined by triangle `tri`.
+Note that the sign depends on triangle point ordering and `d=0` only implies `p`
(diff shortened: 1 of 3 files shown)


sjkelly:
For reference I have a few more of these functions here: https://github.com/sjkelly/Descartes.jl/blob/master/src/frep.jl

weymouth:
> For reference I have a few more of these functions here: https://github.com/sjkelly/Descartes.jl/blob/master/src/frep.jl

That's great. Do you think it makes sense to merge some of them?

src/distances.jl
```diff
@@ -0,0 +1,117 @@
...
+signed_distance(p,s::HyperSphere) = norm(p-origin(s))-radius(s)
+
+"""
+    signed_distance(p,r::Rect)
+
+Return the signed distance from `p` to `r`.
+"""
+function signed_distance(p,r::Rect)
+    # reflect p to positive Rect quadrant and get vector relative to Rect corner
+    @show p,origin(r),width(r)
```
weymouth, line 113: That shouldn't be there...

sjkelly:
Yes, feel free copy whatever is useful from Descartes, assuming people are on board with adding these distance functions.

SimonDanisch:
Thank you :) 
Yeah I don't see why not ;)

weymouth:
So the three primitives I haven't covered yet are `Particle` (trivial) `Cylinder` and `Pyramid`. 

It looks like `Cylinder` is defined with a general line segment and a radius (this is different than Descartes, so I can't copy the code over). I can generalize the `Rect` corner distance function to cover these.

A `Pyramid` is just a union of planes, so it'll be caught with a fallback of meshing anything `meshable`.

src/distances.jl
```diff
@@ -0,0 +1,117 @@
...
+signed_distance(p,tri::Triangle) = signed_distance_to_tri(p,tri.points.data...)
+signed_distance_to_tri(p,a,b,c) = normalize(orthogonal_vector(a,b,c))⋅(p-a)
+
+nonzero(mesh::AbstractMesh) = (i for i in mesh if sum(abs2,orthogonal_vector(i.points.data...))>0)
+"""
+    absolute_distance(p, mesh::AbstractMesh)
+
+Return the minimum absolute distance from point `p` to any Triangle in `mesh`.
+"""
+absolute_distance(p,mesh::AbstractMesh) = minimum(t->absolute_distance(p,t), nonzero(mesh))
```
sjkelly, line 82: This is going to scale poorly on performance, so I'm not sure if this will give the right experience. IIUC for Mesh->SDF you normally want to bin faces in an octree, kd tree, or other, to try to reduce the search space first.

sjkelly:
I'm on the fence about distances for `AbstractMesh`. There are techniques to speed this up, and what is in here is of bad search complexity for anything decently large. Moreso if you want to do uniform sampling of the mesh. To do it fast, we would need some other dependencies, that would introduce a circular dependents. I'm a little out of the loop on invalidations and how they would affect the design decision here as well.

src/distances.jl
weymouth: Very true. This is where some help from people more familiar with these kind of approaches would be great.

weymouth:
By uniform sampling, do you mean filling an entire array with the SDF of a mesh? Because that _is_ my use case.

In fact, I only need this to be accurate fairly close to the mesh, so my plan was to loop through all the mesh faces and fill the array points in the face's bounding box. So this would scale with the number of faces times number of array points per face. 

For the general case, I agree that you need some kind of tree method, but this requires some data-structures be set up within GeometryBasics. You could also apply a tree to the array, such as a multi-grid search.

sjkelly:
Nice! This would be useful to a few other packages, as-is so I am for it.
I realize too you could integrate with [Adaptive Distance Fields](https://juliahub.com/ui/Packages/AdaptiveDistanceFields) and address a good chunk of performance issues. Also Interpolations.jl would help in many cases.

weymouth:
Hold the phone. Looking at `AdaptiveDistanceFields` (which looks cool) `EnhancedGJK` already has a function `ReferenceDistance.signed_distance` to a mesh of simplexes, which means I don't need to implement it again here.

weymouth:
But `EnhancedGJK` only works with `GeometryTypes`!!?!

SimonDanisch:
>But EnhancedGJK only works with GeometryTypes!!?!

It's usually not too hard to upgrade to GeometryBasics, since most APIs have stayed fairly similar!

weymouth:
What is the process/etiquette for this? I've raised an issue on `EnhancedGJK` but not heard back. https://github.com/JuliaRobotics/EnhancedGJK.jl/issues/39

SimonDanisch:
Good question... I think those may not be maintained anymore...  We can ask, if they want to move the package to JuliaGeometry, so it can be maintained by more people!

sjkelly:
FWIW the GT->GB transition is a pretty big ask. A huge amount of their stack is in GeometryTypes, with EnhancedGJK being pretty key foundation. One approach would be to use `import` rather than `using` and support both APIs in EnhancedGJK. I did this in Meshing.jl because many of my users were still on GeometryTypes. 
https://github.com/JuliaGeometry/Meshing.jl/blob/master/src/geometrybasics_api.jl
https://github.com/JuliaGeometry/Meshing.jl/blob/master/src/geometrytypes_api.jl

PR #165, opened 2022-01-22
remove generated Rect method

piever:
In recent julia versions, generated functions aren't really needed for good code generation. For example, the method below generates this

```julia
julia> @code_typed Rect(1, 2, 3, 4, 5, 6, 7, 8)
CodeInfo(
1 ─       nothing::Nothing
│   %2  = Base.getfield(vals, 1, true)::Int64
│   %3  = Base.getfield(vals, 2, true)::Int64
│   %4  = Base.getfield(vals, 3, true)::Int64
│   %5  = Base.getfield(vals, 4, true)::Int64
│   %6  = Base.getfield(vals, 5, true)::Int64
│   %7  = Base.getfield(vals, 6, true)::Int64
│   %8  = Base.getfield(vals, 7, true)::Int64
│   %9  = Base.getfield(vals, 8, true)::Int64
│   %10 = Core.tuple(%2, %3, %4, %5)::NTuple{4, Int64}
│   %11 = %new(Vec4{Int64}, %10)::Vec4{Int64}
│   %12 = Core.tuple(%6, %7, %8, %9)::NTuple{4, Int64}
│   %13 = %new(Vec4{Int64}, %12)::Vec4{Int64}
│   %14 = %new(Recti{4}, %11, %13)::Recti{4}
└──       return %14
) => Recti{4}
```

I'm planning to see if there are other generated functions that can removed easily (not sure if in this PR or a separate one).

EDIT: I'm a bit confused as to why tests are failings, but they seem to fail also on master locally for me, not sure what's wrong.


sjkelly:
This seems good to go?

piever:
@SimonDanisch was commenting (in a separate discussion) that in his experience generated functions can actually be faster to compile than the non-generated counterpart and was afraid that this could slow down compile times, so maybe it's better to spend some time deciding whether we want this or not.

I'm actually not completely sure what's the cleanest way to benchmark whether this has a positive or negative impact in terms of compile time.

PR #166, opened 2022-02-02
simplify snip signature

piever:
As this methods seemed very costly in https://github.com/JuliaPlots/Makie.jl/discussions/1636, I thought it could be less taxing on the compiler to remove unused type parameters.

### DIFF
--- a/src/triangulation.jl
+++ b/src/triangulation.jl
@@ -85,7 +85,7 @@ function Base.in(P::T, triangle::Triangle) where {T<:AbstractPoint}
     return ((a_bp >= t0) && (b_cp >= t0) && (c_ap >= t0))
 end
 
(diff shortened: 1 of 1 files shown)


sjkelly:
Seems reasonable. Though the CI seems to have bitrotted? @piever I notice you are the lead dev for StructArrays.jl. Any insight on this? https://github.com/JuliaGeometry/GeometryBasics.jl/runs/5034017528?check_suite_focus=true

piever:
Sure, I've added #167 that fixes that (it was some usage of internals of StructArrays), even though that still does not fix CI (see [here](https://github.com/JuliaGeometry/GeometryBasics.jl/runs/5042660294?check_suite_focus=true#step:6:95)), but that issue at least seems unrelated to StructArrays.

sjkelly:
Ah thanks for the force push :)

PR #167, opened 2022-02-02
fix meta_table

piever:
GeometryBasics was using an internal function of StructArrays that was changed during a refactor, this PR avoids depending on internals of StructArrays.


sjkelly:
Thanks! CI Failures seem to be isolated to 1.7.1. 1.6 tests fine for me locally, though I see the same CI failure on 1.7 locally. We should bump anyway since 1.6 is the new LTS.

PR #169, opened 2022-02-25
Get rid of StaticArrays

SimonDanisch:
And some other type clean ups.


sjkelly:
This doesn't seem like the correct long term solution if we can cache precompiles between modules in 1.8+. 

Xref discussion from a while back: https://github.com/JuliaGeometry/GeometryBasics.jl/issues/91

SimonDanisch:
> if we can cache precompiles between modules in 1.8+.

That's not helping, since StaticArrays emits much harder code to compile...So until we cache binary, this is a valid approach.
I'm keeping the interface mainly the same, and this also helps remove complex code to wade through while optimizing Makie's compile times - while still allowing us to switch back to using StaticArrays once it makes sense.

Benchmarks indicating the utility of this PR are done with Julia already caching between modules. Since the new code is much less complicated it does indeed shave off around 1-3s.

See: https://github.com/JuliaPlots/Makie.jl/pull/1701
But yes, this opens up the discussion from https://github.com/JuliaGeometry/GeometryBasics.jl/issues/91. Since I'm touching all the Vec code, I'm willing to overthink the Vec/Point interface :)

sjkelly (approved):

sjkelly:
This does share some time off `Plots` also:
```
julia> @time using Plots
  3.701392 seconds (6.20 M allocations: 496.627 MiB, 4.59% gc time, 21.38% compilation time)
```

```
julia> @time using Plots
  3.452578 seconds (6.12 M allocations: 487.336 MiB, 4.05% gc time, 22.26% compilation time)
```

src/fixed_arrays.jl
```diff
@@ -1,127 +1,237 @@
...
-    tup = ntuple(Val(length(T))) do j
-        return ifelse(i == j, 1, 0)
-    end
-    return T(tup)
-end
+using LinearAlgebra
+import Random
+import Base: setindex
+
+abstract type StaticVector{N, T} end
```
hyrodium, line 6: We now have [StaticArraysCore.jl](https://github.com/JuliaArrays/StaticArraysCore.jl), so this definition can be removed.

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.4.2"
 EarCut_jll = "5ae413db-bbd1-5e63-b57d-d24a61df00f5"
 IterTools = "c8e1da08-722c-5040-9ed9-7db0dc04731e"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
```
hyrodium, line 11:
```suggestion
StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
```

src/GeometryBasics.jl
```diff
@@ -1,10 +1,12 @@
 module GeometryBasics
 
-using StaticArrays, Tables, StructArrays, IterTools, LinearAlgebra
+using Tables, StructArrays, IterTools, LinearAlgebra
```
hyrodium, line 3:
```suggestion
using StaticArraysCore, Tables, StructArrays, IterTools, LinearAlgebra
```

hyrodium (commented):
Added some suggestions around StaticArraysCore.jl

Issue #158, opened 2022-01-04
Meshing a single triangle sometimes returns an empty mesh


ikirill:
The doc page at https://juliageometry.github.io/GeometryBasics.jl/dev/meshes/ is empty, so I'm not sure what I'm expected to do, but creating a mesh out of a single triangle sometimes works:

```
julia> normal_mesh(Triangle(Point(0,0,0.0),Point(1.0,0,0),Point(0,1.0,0)))
Mesh{3, Float32, Triangle}:
 Triangle(Float32[0.0, 1.0, 0.0], Float32[0.0, 0.0, 0.0], Float32[1.0, 0.0, 0.0])
```

and sometimes doesn't:

```
julia> ϕ = (sqrt(5)+1)/2
1.618033988749895
julia> p,q,r = Point(ϕ,0,+1),Point(1,ϕ,0),Point(ϕ,0,-1)
([1.618033988749895, 0.0, 1.0], [1.0, 1.618033988749895, 0.0], [1.618033988749895, 0.0, -1.0])
julia> normal_mesh(Triangle(p,q,r))
Mesh{3, Float64, Triangle}
```

I was trying to follow the MeshCat.jl's tutorial, and I got this issue while trying to draw a flat triangle.

sjkelly:
Can you link the MeshCat.jl tutorial? I think MeshCat.jl is on GeometryTypes :/

SimonDanisch:
I debugged this a bit, it's some weirdness in this constructor:
https://github.com/JuliaGeometry/GeometryBasics.jl/blob/master/src/meshes.jl#L91
faces returns nothing, and therefore it tries to triangulate the triangle 😅  I guess the fix would be to implement `faces(::Triangle)` to return `[GLTriangleFace(1, 2, 3)]` (or 0, 1, 2)...

Issue #151, opened 2021-08-11
Ambiguous NgonFace constructors


lhupe:
Constructing a `NgonFace` from an `SVector` fails because of ambiguity: `StaticArrays` defines some conversion constructors, which interfere with the `NgonFace` constructors because `NgonFace <: StaticVector`

## MWE 

```julia
using GeometryBasic, StaticArrays
t = TriangleFace(SA[0.4, 0.2, 0.55])
```
results in 
```julia
ERROR: MethodError: (TriangleFace{T} where T)(::SVector{3, Float64}) is ambiguous. Candidates:
  (::Type{SV})(x::StaticVector{N, T} where {N, T}) where SV<:NgonFace in GeometryBasics at /home/$USER/.julia/packages/GeometryBasics/WMp6v/src/fixed_arrays.jl:74
  (::Type{SA})(a::StaticArray) where SA<:StaticArray in StaticArrays at /home/$USER/.julia/packages/StaticArrays/vxjOO/src/convert.jl:5
  (NgonFace{S, T} where T)(x::AbstractVector{T}) where {S, T} in GeometryBasics at /home/$USER/.julia/packages/GeometryBasics/WMp6v/src/fixed_arrays.jl:34
  (::Type{SA})(a::AbstractArray) where SA<:StaticArray in StaticArrays at /home/$USER/.julia/packages/StaticArrays/vxjOO/src/convert.jl:7
  (NgonFace{S, T} where T)(x::T) where {S, T} in GeometryBasics at /home/$USER/.julia/packages/GeometryBasics/WMp6v/src/fixed_arrays.jl:48
Possible fix, define
  (NgonFace{S, T} where T)(::StaticVector{N, T} where N) where {S, T}
Stacktrace:
 [1] top-level scope
   @ REPL[4]:1
```

Issue #142, opened 2021-03-27
Mesh with tris and quads?


RibeiroAndre:
Hi!
Any way to create a mesh of mixed element types? I don't see any way of mixing tris and quads in the same mesh.
Thanks a lot!

juliohm:
Take a look at Meshes.jl for mixed element meshes. We have a simple
UnstructuredMesh representation and will be adding other data structures
soon for fast navigation over the faces of any kind.

RibeiroAndre:
Thanks for the quick answer.
Would the meshes from Meshes.jl be compatible with Makie?
Thanks!

juliohm:
Not yet but we are working on recipes and trying to provide feedback to
improve the Makie recipe system.

RibeiroAndre:
Ok, thanks!

SimonDanisch:
There is an overload missing to create the correct polytope type:
```julia
using GeometryBasics
using GeometryBasics: AbstractNgonFace, AbstractPoint, Ngon
function GeometryBasics.Polytope(P::Type{<:AbstractPoint{Dim,T}},
                  ::Type{<:AbstractNgonFace{N,IT} where N}) where {Dim,T,IT}
    return Ngon{Dim,T,N,P} where N
end

f = [TriangleFace(1, 2, 3), QuadFace(1, 2, 3, 4)]

p = Point2f0[(0, 1), (1, 2), (3, 4), (4, 5)]
Mesh(p, x);
```
There seem to be some more overloads missing for when there are faces with different Ns, so display errors as well...

BambOoxX:
Hi @SimonDanisch , I just stumbled upon this very issue, is there a plan to implement a correction for this issue ? 

I also tried your fix (just correcting `x` by `f` in `Mesh(p,x)` and it shows no error as long as the output is omitted with `;` as you proposed. 
However, if i remove the trailing `;` an exception occurs
```
ERROR: MethodError: no method matching (Ngon{2, Float32, N, Point{2, Float32}} where N)(::QuadFace{Point{2, Float32}})
```

I'm working with 
- GeometryBasics v0.4.1
- Julia v1.7.1

ffreyer:
Fixed in 0,.5

Issue #136, opened 2021-02-27
merging an empty vector of meshes / creating an empty mesh


jkrumbiegel:
This issue https://github.com/JuliaPlots/Makie.jl/issues/859 happens when an empty vector of meshes is merged. Currently, this errors, but I think it should create an empty Mesh. I don't know if that's currently possible, but I guess it should be, as these edge cases with zero objects happen all the time in dynamic plotting. It should be possible to draw an empty mesh

ffreyer:
Fixed in 0.5

PR #173, opened 2022-03-17
Slim down GeometryBasics and remove all type complexity

SimonDanisch:
So... I tried for a very long time to create a Geometry package that serves all use cases and can be used as a foundation for all other Geometry packages without any performance problems for any domain. This never really worked out for a bunch of reasons.
While GeometryBasics was never used that way, all the ideal requirements I aimed for have turned GeometryBasics into a very overengineered package, with very long compile times and very complex types.
I never really had a lot of time to actually maintain and grow it though, so on top of being overly complex it has been poorly documented. I also tried to integrate lots of types like LineStrings, which I don't know very well and when to use them :D
So, this PR is slimming down GeometryBasics to the very basics that I need in Makie, which should help a lot with compile times and maintainability.

It changes:
* removes all kind of meta infrastructure, only leaving a simple MetaMesh type that can hold e.g. uv and normals
* remove table interface
* less type artistry and instead tries to use the most simple types as possible
* copies more instead of creating horrible types like `StructArray{ReinterpretArray{SubArray{Int64, 1, Vector{Int64}, Tuple{UnitRange{Int64}}, true}}}` (this is even shorted :D ) 
* Make Point +/-/* Vec consistent

Fixes:
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/158
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/151
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/142
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/136
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/95
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/81
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/60
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/13

Will not be in the scope of this package anymore:
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/71 
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/50
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/49
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/29
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/31
* https://github.com/JuliaGeometry/GeometryBasics.jl/issues/26


evetion:
This is some good cleanup! This is on top of #169 right?

Over at https://github.com/JuliaGeo/GeoInterface.jl/pull/33, we're now (finally) finishing a new Traits based interface. This gets rid of all the required subtyping on GeoInterface.Geometries and I would like to update the GeoInterface implementation in GeometryBasics here as well.

In that context, I see you removed `LineString` in this PR and changed the decomposition in this PR. In GeoInterface we still assume a Polygon decomposes into 1 or multiple LineStrings like instances (exterior ring and any interiors), which themselves decompose into Pointlikes. Do you think we could still make that work in this PR? Would `Line` be a proper replacement, or is it just two points?

SimonDanisch:
Yes, I think we can bring it back, if we document a bit better what its supposed to be used for!
I think we should keep them as simple as possible, so maybe something like:
```julia
struct LineString{N, T}
   line::Vector{Point{N, T}}
end

struct MultiLineString{N, T}
   lines::Vector{LineString{N, T}}
end
```

evetion:
@SimonDanisch Is it ok, to revive https://github.com/JuliaGeometry/GeometryBasics.jl/compare/master...visr:rfc?expand=1 on the current master of GeometryBasics? Or would you like me to open a PR, based on the changes in this PR? Also depends on your expected timeline of this PR?

SimonDanisch:
No, that sounds great :) Looks like it has 0 dependencies, so that makes things a lot easier ;) 
I think master is fine... It's such a small diff, that it should be easy enough to rebase in this PR.

evetion:
> No, that sounds great :) Looks like it has 0 dependencies, so that makes things a lot easier ;) I think master is fine... It's such a small diff, that it should be easy enough to rebase in this PR.

Ok 👍🏻 , I've made a first attempt over at #175.

jeremiahpslewis:
@SimonDanisch Is the PR still active? From its description, it would solve some issues with adding GeoInterface support to GeoMakie.

SimonDanisch:
Yeah, still planning to merge it! 
Need to first make sure that Makie works though, which is a work in progress: https://github.com/JuliaPlots/Makie.jl/pull/2220

termi-official (commented):
I think the changes here are a major improvement of the package! It is really way easier to follow what happens now. However, I still have some questions here.
1. What are the operational and semantic difference between AbstractGeometry and GeometryPrimitive? My understanding on the semantics is that the latter defines "atomic geometry", while the former also includes composites like the meshes. Is this correct or is there more?
2. What is the minimal example to make a new data type interoperable with Makie? Do we even need the types above for this or is it sufficient to provide a fixed set of dispatches?
3. Why do the polygons carry the points (in contrast to the mesh carrying a point vector and the polygons carry indices to this vector)?

SimonDanisch:
1. Good question, I think I introduced that so that one can differentiate between primitives in GeometryBasics and outside... It's been a while, but I think I expected other packages to directly inherit from AbstractGeometry, or their own abstract subtype.
2. It depends a lot on the type - there isn't a clear interface. Maybe we should have one though.
3. In OpenGL you almost always have faces + points, so for this refactor I simplified it to always have the indices. For Polygons you shouldn't share points that much... I think a Polygon quickly becomes invalid if you share points?
4. There isn't really any interaction beyond that MatNf + VecNf can be directly used as Uniforms via GLAbstraction.

rafaqz:
Yes for 3, in most other geometry frameworks points are embedded in the polygon object. 

Accessing indexed points is slower due to the indirection, and needs ~50% more memory in the (2d) case that no points are shared. It would only occasionally save ~25% memory when all points are shared.

Kevin-Mattheus-Moerman:
@SimonDanisch @termi-official @rafaqz would it be worth resuming this discussion or has this gone stale? I am a new user but heavily invested now in GeometryBasics so would welcome and improvements like discussed here. 

I can add to the discussion on this point by @termi-official: 
> Why do the polygons carry the points (in contrast to the mesh carrying a point vector and the polygons carry indices to this vector)? 

I can see @rafaqz 's point that in some cases not shipping the points is inefficient (indexing takes time). But in computational methods (physics, engineering, finite element analysis) and geometry processing, this is really annoying and makes coding very inefficient. For instance shared nodes should be shared. If each face/element has its own points it may be convenient for rendering/visualisation, but when adjacent entities should share nodes this approach creates a excess of points that are really not needed. And this would also separate the entities that should really be attached (share points)*. This causes users to have to merge and split points repeatedly which makes every processing step very inefficient. Below I illustrate this problem with the help of a helpless bunny rabbit, left is imported STL (where each face has its own points), on the right after merging points are properly shared). 
So it sounds like we have users/requirements for both of these approaches (1) just indices, 2) ship with points) so perhaps GeometryBasics should support both? 

https://github.com/JuliaGeometry/GeometryBasics.jl/assets/8392709/402e0146-93c5-49ea-ac0d-63fd7b72368d

For context, I am the developer for [Comodo](https://github.com/COMODO-research/Comodo.jl) and just posted this related issue (before finding this PR): https://github.com/JuliaGeometry/GeometryBasics.jl/issues/217

ffreyer:
Let me add some wishes/TODOs:

An obj file can reference an mtl file which contains material information such as lighting coefficients and paths to texture images (possibly including normals maps, specular maps, etc). These are then referenced by the obj file for groups of faces. To support that we would need:
- non-vertex meta data (maybe just a Dict/NamedTuple)
- sub-meshes or face groups (I think just ranges into faces would work here, maybe via non-vertex meta data)
- maybe a type to contain texture meta data (origin, scaling, channel - check options of object file for this?)

If we want to support skeletal animation at some point we need:
- a bone-tree so we know how the bone matrices depend on each other
- per vertex bone indices and weights (probably just a vec4 each)
- maybe a cache for bone transformation matrices

Some other thoughts:
- It would be nice if 2D primitives could have 3D origins, e.g. to place a 2D ground plane in 3D space
- If there is a simple way to enable Meshes.jl to use GeometryBasics we should probably include it. Perhaps making position types parametric in mesh types would be enough?

ffreyer:
Another thing I forgot to mention is the issue of vertex duplication to get flat shading. E.g. that Rect is duplicating every vertex 3 times so that the corresponding normal can be different for each connected QuadFace. The way we are handling this atm means that even without normals, positions get duplicated. That has been annoying me with bounding boxes in Makie and I assume that is also an annoyance for Geometry related work. (Related #139)

A potential fix for this (specifically avoiding position duplication when vertex duplication is necessary) could be to allow `faces()` to output different indices for positions, normals and uvs. E.g.:
```julia
function faces(obj::Cylinder)
    pos_faces = # generate for position array
    normal_faces = # generate for normals array
    uv_faces = pos_faces
    return pos_faces, normal_faces, uv_faces
end

function create_normal_mesh(primitive)
    ps = coordinates(primitive)
    ns = normals(primitives)
    coord_fs, normal_fs, uv_fs = faces(primitive)
    if coord_fs === normal_fs
        return Mesh(ps, coord_fs; normals = ns)
    else
        fs, mappings = merge_faces(coord_fs, normal_fs) # remap with duplication
        ps = ps[mappings[1]]
        ns = ns[mappings[2]]
        return Mesh(ps, fs; normals = ns)
    end
end
```
where `merge_faces` would be this function that's currently sitting in MeshIO: https://github.com/JuliaIO/MeshIO.jl/blob/cb8e494a61a7b4458cbe6274ac75cf4ba91513e1/src/util.jl#L13. This way we could have meshes with only positions as well as `coordinates()` and `texturecoordinates()` skip the duplication, but generating a normal_mesh would be more expensive.

ffreyer:
Regarding the `Mesh` type(s), bundling vertex attributes with a "basic" mesh and sub-mesh/global attributes with a more complex wrapper seems reasonable to me. E.g.
```julia
struct Mesh{
        # Dim, ElType <: Real, # are these useful to expose?
        PointType <: AbstractPoint, # allow other Point types
        PointCollection <: AbstractArray{PointType}, # maybe to allow GPU Arrays?
        FaceType <: AbstractFace
    }
    positions::PointCollection
    faces::Vector{FaceType}
    # per vertex data, e.g. normals, uvs, colors, ...
    vertex_meta::Dict{Symbol, AbstractVector}
end

struct MetaMesh{..., MT <: Mesh{...}} <: AbstractMesh
    # maybe a merged mesh from multiple submeshes
    mesh::MT
    
    # index ranges for mesh.faces which correspond to
    # one submesh
    submesh_views::Vector{UnitRange{UInt32}}
    
    # non-vertex data, either per submesh or fully global
    meta::Dict{Symbol, Any}
end
```
I don't think it's necessary to separate per-submesh (e.g. diffuse multiplier) and global data (e.g. a texture image) because I currently imagine these would be handled rather explicitly. For rendering I think it's optimal to upload all the vertices and faces once and then draw different sections by specifying different start pointers + counts, updating the necessary uniforms in between.

termi-official:
> If we want to support skeletal animation at some point we need:
> 
>     * a bone-tree so we know how the bone matrices depend on each other
> 
>     * per vertex bone indices and weights (probably just a vec4 each)
> 
>     * maybe a cache for bone transformation matrices
> 

Really like the idea of having functionality for skeletal animation in Julia. However, should this really go into GeometryBasics.jl? I would have rather thought that this is something for a separate package.

ffreyer:
I only skimmed the learnopengl tutorial so far, so my understanding might be a bit off. But I think a good chunk of the data is closely related to a mesh. Weights and bone indices are vertex attributes, which would already be supported by the current GeometryBasics. 
The bone tree gives context to the bone indices and describes how the bone transforms inherit from each other. Data-wise I think it's debatable if that should be part of a mesh. My main motivation for wanting compatibility for it is that MeshIO may eventually be able to load it and I'd like it to still just return a mesh. 
In terms of functionality I don't think GeometryBasics should do much. I really just want it to be able hold all the data that mesh file formats store so they can be used somewhere else.

SimonDanisch:
I'm going to merge this, to clean up pending pull requests etc.
This doesn't mean it's 100% ready, I expect a bit more polish for the docs, and maybe a few more little clean ups once we use it more.

PR #175, opened 2022-06-12
Implemented GeoInterface.

evetion:
- [x] Add some extra tests
- [x] ~Slim down implementation/types~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - master
-    tags: '*'
+    tags: "*"
   pull_request:
     branches:
       - master
@@ -15,8 +15,8 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.3'
-          - '1.7'
+          - "1.6"
+          - "1"
         os:
           - ubuntu-latest
           - macOS-latest
@@ -54,7 +54,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: julia-actions/setup-julia@v1
         with:
-          version: '1.7'
+          version: "1.7"
       - run: |
           julia --project=docs -e '
             using Pkg
--- a/Project.toml
+++ b/Project.toml
@@ -23,9 +23,10 @@ julia = "1.6"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
+GeoJSON = "61d90e0f-e114-555e-ac52-39dfb47a3ef9"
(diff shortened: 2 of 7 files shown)


evetion:
@sjkelly Is it ok if we bump the minimum required Julia version to 1.6 here?

visr:
Makie also still has julia 1.3 compat I see (though it seems CI from 1.6). Why would this be breaking by the way? It just adds a feature right?

evetion:
Well the GeoInterface is a breaking release, so what worked in the old version won't work in the new one. I'd say it's breaking, but it depends on what you consider part of the package.

visr:
I'd say since the GeoInterface is only added now that it is v1 it is not breaking, for GeometryBasics.
Bumping the Julia compat to 1.6 also does not have to be a breaking release these days, right? Since older (unsupported) julia version will just not pick it up

sjkelly:
I am not opposed to dropping pre-1.6. @SimonDanisch does the next version of Makie targeting 1.3 compat?

SimonDanisch:
I think we're slowly giving up on 1.3, so fine by me I suppose!

sjkelly (approved):

sjkelly (approved):

sjkelly:
@SimonDanisch do you want to check this for load time regressions since it adds a few new dependencies?

SimonDanisch:
Huh, I guess those aren't due to this PR:
image

But seems like we still need compat bounds for GeoInterface, though!


>do you want to check this for load time regressions since it adds a few new dependencies?

I just tried, and if there's one it's barely measurable... Which makes sense, since the dependency is very small, so shouldn't be in the way ;)

visr:
For the deprecation warnings, they come from `Aqua.test_all`, I just filed https://github.com/JuliaTesting/Aqua.jl/issues/83.

> But seems like we still need compat bounds for GeoInterface, though!

I'll add those plus a few Base.convert methods that will allow us to convert geometry types through the interface, https://juliageo.org/GeoInterface.jl/dev/guides/developer/#Conversion

visr:
Ok I finished the remaining items, this is good to go.

lazarusA:
wow! @visr rocks!!!

visr:
Haha many have worked on this in some way. But I'm glad it's merged, it will definitely help the interoperability of the ecosystems.

```julia
using GADM, CairoMakie, GeoInterface
using CairoMakie: MultiPolygon

gdalgeom = only(GADM.get("UKR").geom)
basicgeom = GeoInterface.convert(MultiPolygon, gdalgeom)
poly(basicgeom)
```

download (Small)

PR #180, opened 2022-09-16
use concrete points

SimonDanisch:
@visr @evetion is there a better way to figure out the eltype?


SimonDanisch:
I want this tagged and can't reproduce the CI failure...Will try later, meanwhile, I'm pretty sure that the CI is not failing because of this PR

visr:
I think this is fine. Float64 is the most common coordinate type. There was some related discussion in https://github.com/JuliaGeo/GeoInterface.jl/issues/53.

PR #181, opened 2022-10-08
fix Rect3 uv coordinates

ffreyer:
The current uv's in Rect3 are just the xy components of the coordinates. That means uvs are doubly assigned and not normalized. This pr adds some manual uvs so that
```julia
img = [
    RGBf(1,0,0) RGBf(0,1,0) RGBf(0,0,1);
    RGBf(0,1,1) RGBf(1,0,1) RGBf(1,1,0);
]
```
colors a cube. (Colors applying to the `[+x, +y, +z; -x, -y, -z]` face.)

### DIFF
--- a/src/primitives/rectangles.jl
+++ b/src/primitives/rectangles.jl
@@ -553,7 +553,17 @@ function coordinates(rect::Rect3)
 end
 
 function texturecoordinates(rect::Rect3)
-    return coordinates(Rect3(0, 0, 0, 1, 1, 1))
+    # compatible with a texture ordered:
+    # [+x +y +z; -x -y -z]
+    uvs = map(v -> v ./ (3, 2), Vec{2, Float32}[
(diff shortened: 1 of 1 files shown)


SimonDanisch:
> and not normalized.

They should be since I'm using the unit cube, no?
Also, they need to return Vec3s, to work with 3d uvw coordinates...

ffreyer:
> They should be since I'm using the unit cube, no?

Oh, you're right.

> Also, they need to return Vec3s, to work with 3d uvw coordinates...

Why do we need 3d uvw coordinates though? From what I've seen working with 2d uv coordinates is much more common. Makie also doesn't accept 3d textures, e.g. `mesh(Rect3f(Point3f(0), Vec3f(1)), color = rand(RGBf, 2, 2, 2))` errors. So you can't use them there.

ffreyer:
Isn't it much more common to texture a surface than a volume though?

PR #185, opened 2022-12-06
Fix #168

### LINKED ISSUE #168: Bug in Line Intersection ( GeometryBasics.intersects )
I found a pair of lines that `GeometryBasics.intersects` claims don't intersect even though they do.  How to reproduce:

I'm using `Julia 1.7.2` with `[5c1252a2] GeometryBasics v0.4.1`

If you run the following code
```
using GeometryBasics
const GB = GeometryBasics
mkline(pair) = GB.Line(GB.Point(pair[1]...), GB.Point(pair[2]...))
bug2 =(mkline([-3.1, 15.588457268119894] => [3.1, 15.588457268119894]),
    mkline([2.0866025403784354, 17.37050807568877] => [-4.0866025403784505, 13.806406460551015]))
GB.intersects(bug2[1], bug2[2])
```
I get `(false, [0.0, 0.0])` even though these lines should intersect.  To confirm that they do I've drawn the lines using the `Luxor` package here: https://imgur.com/a/3sADDln

Drawing code for ease of copy paste:
```
using Luxor
@draw begin
    LP = Luxor.Point
    dscale=10
    setline(1)
    sethue("black")
    for (p1, p2) in bug2
        asp(p) = let 
            v = ([dscale, -dscale] .* vec(p))
            @show v
            Luxor.Point(v...)
        end
        line(asp(p1), asp(p2); action=:stroke)
    end
end
```

Anyway, thanks for having made the geometry package, and I'll be curious to see what the bug was.

cmcaine:
Replaces the intersection algorithm with one that seems more reliable.

See here for the algorithm and explanation: http://paulbourke.net/geometry/pointlineplane/

This algorithm is also used in Turf.js, Turf.jl, etc.

I did a bunch of testing with other algorithms and even tested some algorithms from ChatGPT. You can see some of the testing here: https://gist.github.com/cmcaine/80600c1fd32a6ac3822f2fac4ab93367

### DIFF
--- a/src/lines.jl
+++ b/src/lines.jl
@@ -2,66 +2,47 @@
 """
     intersects(a::Line, b::Line) -> Bool, Point
 
-Intersection of 2 line segmens `a` and `b`.
-Returns intersection_found::Bool, intersection_point::Point
+Intersection of 2 line segments `a` and `b`.
+Returns `(intersection_found::Bool, intersection_point::Point)`
 """
+# 2D Line-segment intersection algorithm by Paul Bourke and many others.
+# http://paulbourke.net/geometry/pointlineplane/
 function intersects(a::Line{2,T1}, b::Line{2,T2}) where {T1,T2}
     T = promote_type(T1, T2)
-    v1, v2 = a
-    v3, v4 = b
-    MT = Mat{2,2,T,4}
     p0 = zero(Point2{T})
 
-    verticalA = v1[1] == v2[1]
-    verticalB = v3[1] == v4[1]
+    x1, y1 = a[1]
+    x2, y2 = a[2]
+    x3, y3 = b[1]
+    x4, y4 = b[2]
 
-    # if a segment is vertical the linear algebra might have trouble
-    # so we will rotate the segments such that neither is vertical
-    dorotation = verticalA || verticalB
+    denominator = ((y4 - y3) * (x2 - x1)) - ((x4 - x3) * (y2 - y1))
+    numerator_a = ((x4 - x3) * (y1 - y3)) - ((y4 - y3) * (x1 - x3))
+    numerator_b = ((x2 - x1) * (y1 - y3)) - ((y2 - y1) * (x1 - x3))
 
-    if dorotation
-        θ = T(0.0)
-        if verticalA && verticalB
-            θ = T(π / 4)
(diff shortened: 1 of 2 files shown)


rafaqz (commented):
This looks good to me. Have you done a benchmark comparison?

cmcaine:
Thanks for the review! I probably benchmarked it at the time but I've forgotten now.

It's a much simpler algorithm and I would be very surprised if it was slower than the existing code.

For me, the more important thing is that this is a well-known and peer-reviewed algorithm and the current implementation is not (perhaps it's a published algorithm, but I'm not familiar with it and it's provenance isn't noted in the code). The Paul Bourke algorithm I've used also seems to give better quality results, as evidenced by the tests I added, etc.

cmcaine:
Oh, there are benchmarks and a bunch of code evaluating the accuracy of various line intersection algorithms in my gist that I linked in my first post.

https://gist.github.com/cmcaine/80600c1fd32a6ac3822f2fac4ab93367

rafaqz:
I mean, posting the `btime` output here is the path to getting this merged, no-one will read your link.

I dont have commit access here.

cmcaine:
I'm not gonna rerun this stuff for a package I don't use any more and a PR that I don't believe is ever going to be merged. 🤷

The short of it is that the algorithm in this PR is better known, better tested, seems to produce better output and (least importantly, imo) is almost certainly faster.

Simon can take it or leave it. I appreciate that they're spread thin.

rafaqz:
No worries! I use GeometryOps.jl for this ;)

It would just be good to not have correctness issues around and proving inline in comments that a PR is all upside no downside is a pretty reliable way to get it merged by people with zero spare time.

SimonDanisch:
Thank you @cmcaine and sorry for taking so long. I'm not familiar with line intersection algorithm and have little time for review.
Thanks everyone for sanity checking this, I'm not familiar with line intersection algorithms, so having a few more eyes on this helps a lot to actually merge it.

cmcaine:
Thanks knuesel!

Thanks very much Simon for your work on this package (which I maybe use indirectly?) and on many other packages that I and others benefit from 💜.

I lost a few hours to this bug and in writing the PR, but you've saved me many more than that over the years :D

PR #186, opened 2023-01-10
Pass input types to mesh

ffreyer:
For https://github.com/MakieOrg/Makie.jl/pull/2573

Adjusts mesh generating functions to keep the input element type around, rather than converting to Float32


SimonDanisch:
So, these kind of methods were written specifically to convert to OpenGL datatypes by default.
If one doesn't want them, the mesh should just be created with `mesh(...)` or `Mesh(...)` directly...

ffreyer:
Sounds like they should be `gl_normal_mesh` etc then... But I can also change the default conversions in Makie I guess

PR #187, opened 2023-01-26
More GeoInterface methods

asinghvi17:
This PR adds the following features:
- [x] Add more methods for Rect construction, since that's basically used as our bbox type.  Calling `Rect` on any geometry should now return a `Rect` which is the bbox of that geometry.
- [x] Implement `GeoInterface.extent` on top of the previous Rect changes.
- [x] Implement `GeometryBasics.geointerface_geomtype`, which GeoInterface picks up to allow other packages' geometries to be easily converted to GeometryBasics geometries (TODO: should we use this behaviour in Makie somehow?)
- [x] Implement a lot of polygon operations like centroid, distance, signed area, etc.  Also define GeoInterface methods backing these.
- [x] Reformat and add small optimizations to src/geointerface.jl.

TODOs still remaining:

- [x] Add a method to find distance which works well everywhere (the current method only checks for distance to nearest vertex)
- [ ] Look through the whole package, find more methods which GeoInterface methods can call
- [ ] Add tests
- [x] Make sure that performance is good! (tested with Polylabel.jl, it was pretty fast).

### DIFF
--- a/src/boundingboxes.jl
+++ b/src/boundingboxes.jl
@@ -1,7 +1,4 @@
-function Rect(geometry::AbstractArray{<:Point{N,T}}) where {N,T}
-    return Rect{N,T}(geometry)
-end
-
+# the basic method
 """
 Construct a HyperRectangle enclosing all points.
 """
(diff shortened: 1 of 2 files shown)


rafaqz:
Can we break this up into smaller PRs so it gets merged sooner? 

Specifically can we add a PR for `geointerface_geomtypes` just so `convert(GeometryBasices, geom)` works?

asinghvi17:
A lot of this has been taken over by GeometryOps!

SimonDanisch:
So we can close this?

PR #189, opened 2023-01-29
Add geointerface_geomtypes as a method

asinghvi17:
Should get GeoInterface.convert(GeometryBasics, ...) working.

cc: @rafaqz


rafaqz:
LGTM. Doesn't need to be work in progress its done :)

Can we get a merge form someone with commit permissions?

visr (approved):

PR #192, opened 2023-04-03
fix geointerface conversion performance

rafaqz:
This PR fixes type instability in GeoInterface.jl conversion methods.

There are still some issues with TupleView, but I think this is internal to GeometryBasics.jl

There is more that could be done with e.g. Polygon and MultiPolygon, but that can wait for #173 as the types are currently too complicated.


visr (approved):
Looks good to me! I added a commit to fix tests that were too strict.

rafaqz:
Thanks! feel free to merge and tag then, I don't have merge rights in this org

rafaqz:
Ok this is rebased to keep the changes not included in #193

PR #193, opened 2023-04-14
vectors are not `isgeometry` material

rafaqz:
Vectors of geometries were defined as geometries here. But that breaks heaps of the interface as they don't actually return a `geomtrait`.

So removing. No tests as its just deleting things that are bugs.

Another tiny change is using `AbstractMesh` in `isgeometry` instead of `Mesh` because `geomtrait` uses `AbstractMesh`.

### DIFF
--- a/src/geointerface.jl
+++ b/src/geointerface.jl
@@ -4,6 +4,12 @@ GeoInterface.isgeometry(::Type{<:AbstractGeometry}) = true
 GeoInterface.isgeometry(::Type{<:AbstractFace}) = true
 GeoInterface.isgeometry(::Type{<:AbstractPoint}) = true
 GeoInterface.isgeometry(::Type{<:AbstractMesh}) = true
+GeoInterface.isgeometry(::Type{<:AbstractLineString}) = true
+GeoInterface.isgeometry(::Type{<:AbstractPolygon}) = true
+GeoInterface.isgeometry(::Type{<:MultiPoint}) = true
+GeoInterface.isgeometry(::Type{<:MultiLineString}) = true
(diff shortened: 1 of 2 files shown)


rafaqz:
@visr or @SimonDanisch please approve my workflow :pleading_face: 

Or maybe add me as a member here or in JuliaGeometry? this wont be the last PR...

visr:
I approved the workflow but I cannot add you myself.

Note that this breaks `isgeometry` for the Multi geometries, which are not AbstractGeometry, but AbstractVector.

```julia
using GeoInterface, GeometryBasics
mp = MultiPoint([Point(3, 1)])
GeoInterface.isgeometry(mp)  # was true, now becomes false
```

rafaqz:
Ahh right it was using AbstractArray as a catch all for those. I'll add them explicitly

visr:
Right, that works.

Tests still fail, now with `UndefVarError: AbstractLineString not defined`

rafaqz:
Yeah sorry I didn't mean to push and trigger that again. 

I ended up having to fix a lot of things for the latest GeoJSON to work, as its now returning `Float32` - but GeometryBasics should be able to use whatever type it gets. The modified convert methods should be faster anyway because of the branches.

I also added a few more missing tests for the interface while I was there, it's turned into a grab bag of things by now...

visr:
Should #192 be closed in favor of this now?

visr (approved):

rafaqz:
Oh right I thought I had dejavu for those bits. Yes this implementation is better anyway.

PR #194, opened 2023-04-28
Hook up vectors of Points and vectors of those to GeoInterface

asinghvi17:
Vector{Point} is a LineString
Vector{Vector{Point}} is a MultiLineString

Need this for the generic datashader thing :)


rafaqz:
Isn't it ambiguous what a vector of points is though? It could be a MultiPoint or a LineString or a LinearRing...

Why not just wrap them before use?

SimonDanisch:
If we implement the bin counting with OpenGL, we could just use all of the normal rendering pipeline

rafaqz:
Also this is directly clashing with my other PR... I was removing vectors completely as geometries :sweat_smile:

SimonDanisch:
I guess we can close this?

asinghvi17:
Yeah I think so

PR #195, opened 2023-05-01
Add conversions for different parameter types of primitives

ffreyer:
Something like this currently errors
```julia
rs = [Rect2(0,0,1,1)]
rs[1] = Rect2f(0,0,1,1)
```
because there are no conversion functions for these types. This pr aims to fix this for https://github.com/MakieOrg/Makie.jl/pull/2573 which would make it more likely for these types to collide.

### DIFF
--- a/src/primitives/pyramids.jl
+++ b/src/primitives/pyramids.jl
@@ -26,4 +26,4 @@ function faces(::Pyramid)
     return (TriangleFace(triangle) for triangle in TupleView{3}(1:18))
 end
 
-Base.convert(::Type{<: Pyramid}, c::Pyramid) = Pyramid(c.middle, c.length, c.width)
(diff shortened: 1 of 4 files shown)


src/primitives/pyramids.jl
```diff
@@ -25,3 +25,5 @@ end
 function faces(::Pyramid)
     return (TriangleFace(triangle) for triangle in TupleView{3}(1:18))
 end
+
+Base.convert(::Type{<: Pyramid}, c::Pyramid) = Pyramid(c.middle, c.length, c.width)
```
SimonDanisch, line 29: Does this actually do a conversion?

ffreyer: Oh, probably not. I updated it to be like the rest.

I'm also not sure if these converts are ok in general. Having `convert(Rect2f, Rect2f(...))` defined seems weird to me.

PR #196, opened 2023-06-13
Add Extents.extent method for Rect2

felixcremer:
This has been pirated in Tyler and should rather live in GeometryBasics.
This would enable us to use this method in PyramidScheme without having to depend on Tyler for example. 

I put the definition into the GeoInterface file because for me the Extents is an extension of the GeoInterface. 

I am not sure, whether it would make sense to also provide the extent functionality also for higher dimensional rectangles or other geometry types, but the Rect2 is needed for the subsetting in Tyler.

@rafaqz

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -5,6 +5,7 @@ version = "0.4.7"
 
 [deps]
 EarCut_jll = "5ae413db-bbd1-5e63-b57d-d24a61df00f5"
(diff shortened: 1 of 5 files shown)


rafaqz:
Thanks! Good to fix that type piracy.

(Also Extends doesnt actually extend GeoInterface, its not really Geo specific)

rafaqz (approved):

felixcremer:
Bump

PR #198, opened 2023-07-10
don't convert if already of correct type

SimonDanisch:
Correct implementation of https://github.com/JuliaGeometry/GeometryBasics.jl/pull/197
cc: @JonasIsensee


JonasIsensee:
We confirmed locally that this works.
You could add a simple
`@test m == triangle_mesh(m)`
to make codecov happy, here

https://github.com/JuliaGeometry/GeometryBasics.jl/blob/48b1ef6a1fe5cf0e877b318437a0a58169030ee3/test/geometrytypes.jl#L76

PR #199, opened 2023-07-11
improve merge(meshes)

JonasIsensee:
Brings allocations down to a fixed number and improves runtime by about x3 in our application.

Note: `reduce(vcat, ..)` plus broadcasting is faster than `mapreduce` for some reason

### DIFF
--- a/src/meshes.jl
+++ b/src/meshes.jl
@@ -228,16 +228,22 @@ function Base.merge(meshes::AbstractVector{<:Mesh})
     elseif length(meshes) == 1
         return meshes[1]
     else
-        m1 = meshes[1]
-        ps = copy(coordinates(m1))
-        fs = copy(faces(m1))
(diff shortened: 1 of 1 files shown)


src/meshes.jl
```diff
@@ -228,17 +228,23 @@ function Base.merge(meshes::AbstractVector{<:Mesh})
     elseif length(meshes) == 1
         return meshes[1]
     else
-        m1 = meshes[1]
-        ps = copy(coordinates(m1))
-        fs = copy(faces(m1))
+        ps = reduce(vcat, coordinates.(meshes))
```
jkrumbiegel, line 231: couldn't this all be written with a single allocated vector for `ps` and `fs`, and writing to that in the loop? We should know all lengths

JonasIsensee: It surely could.
This was just faster than the more elaborate allocate+loop versions I tested.

Issue #61, opened 2020-06-09
Rename Point3f0 and similar to Point3f, etc.


knuesel:
I'd like to make a case against the suffix `0` in types such as `Point3f0`:
* It confuses beginners (see https://discourse.julialang.org/t/makie-scatter-plot-of-vector-of-points/41039)
* It adds visual noise
* It doesn't really make sense: in the syntax `1.2f0`, the `f` denotes `Float32` and `0` is an exponent. `1.2f4` is also a Float32.
* It is unnecessarily different from what people are used to: Point3f in OpenCV, Rhino, Java vecmath, Eigen (Vector3f), probably many more.

Is there a reason to prefer Point3f0?

Sov-trotter:
@knuesel it's discussed here I guess, https://github.com/JuliaGeo/Shapefile.jl/pull/39

pauljurczak:
As one of the beginners in questions, I fully agree. In the C++ world, you will find almost exclusively Point3f, Vector3f, etc. notation. For different element types, you will see: Point3d, Point3i, etc. The first thing I did after seeing `Point3f0` was:
```
julia> Point3f
ERROR: UndefVarError: Point3f not defined
```

SimonDanisch:
I wouldn't mind in general - but I would mind the work to refactor this and properly get rid of it in all dependencies :D

pauljurczak:
How about an official `Point3f` alias?

knuesel:
@SimonDanisch I can imagine it would be a lot of work, especially with all the dependent projects. Of course if the change is desired, it's better to do it as soon as possible. I would be happy to help with the refactoring (I could work on PRs in the coming weeks if there's interest).

jkrumbiegel:
> I would mind the work to refactor this and properly get rid of it in all dependencies :D

I can't imagine this would be so much work, after all, it's a prime candidate for a simple search and replace, no? Of course tagging the versions would be a bit of work. I like the idea of a rename, especially because we would match the conventional naming, and we should do things like this earlier rather than later.

SimonDanisch:
Yeah, its not a crazy amount! I mainly meant to say, that I won't do it in the little time i have, but if someone is willing to, they're free to open a PR to the dependent packages :)

asinghvi17:
I think a depwarn release of GeometryBasics + a replace in dependent packages should be sufficient.  Here's the list of all (direct and indirect) dependent packages:

```
AbstractPlotting
Arena
[...]

PR #208, opened 2024-01-21
add a docstring for `Rect` and its aliases

putianyi889:


asinghvi17:
Looks good to me!

SimonDanisch:
Thank you! I'm thinking to add a float64 alias here as well here.
How should we call it? `Rectf64`?
@jkrumbiegel, @ffreyer

knuesel:
`Rectd` would be nice and consistent with OpenCV, Rhino, Java vecmath and Eigen (spirit of #61)...

SimonDanisch:
Yeah, that was my favorite before people didn't want it for Point... Its a bit confusing because of `d-> dim`, but if people agree on it, I'd be up for it.

jkrumbiegel:
Maybe `Rectd2` is better than `Rect2d`? But then we'd have to change all the `Point2f`s to `Pointf2`s as well

src/primitives/rectangles.jl
```diff
@@ -14,6 +14,21 @@ end
...
+
+A rectangle in N dimensions, formally the caartesian product of intervals. See also [`HyperRectangle`](@ref). Its aliases are
+
+|Dimensions|Generic    |Float32   |Int       |
+|----------|-----------|----------|----------|
+|Any       |`Rect{N,T}`|`Rectf{N}`|`Recti{N}`|
+|2         |`Rect2{T}` |`Rect2f`  |`Rect2i`  |
+|3         |`Rect3{T}` |`Rect3f`  |`Rect3i`  |
+
+and `RectT{T} = Rect{N,T} where {N}`.
```
knuesel, line 28:
```suggestion
|Any       |`Rect`  |`Rectf`   |`Recti` |
|2         |`Rect2` |`Rect2f`  |`Rect2i`|
|3         |`Rect3` |`Rect3f`  |`Rect3i`|

For example `Rect2 == (Rect{2,T} where T)` and `Rectf == (Rect{N,Float32} where N)`.
```

In the table I think it's better to show the aliases themselves, which can be used directly by the user and are less confusing to beginners, rather than the generic parametrized types that can only be used with a `where` clause.

putianyi889: I think you misunderstand what these aliases are, which means this doc is not well written... I'll rephrase them.

knuesel:
> Maybe `Rectd2` is better than `Rect2d`? But then we'd have to change all the `Point2f`s to `Pointf2`s as well

Is the concern that "Rect2d" reads like "2 dimensional rect"? I thought the same but I found it quite easy to get used to `Point2d`, `Vec2d`, `Rect2d`. Some things to consider:

* The `2`,`d` sequence in `Rect2d` corresponds to the sequence of type parameters: it's `Rect{N,T}` not `Rect{T,N}`. It's also natural to have the "smallest information" come last, i.e. Float32 vs Float64 is a smaller change than 2D vs 3D.
* I find easier to read `Rect2f`, `Rect2d`, `Vec2d`, than `Rectf2`, `Rectd2`, `Vecd2`. The digit gives a nice visual separation between the word and the `f` or `d` suffix.

I suspect these two points are why most APIs use `Rect2d` rather than `Rectd2`.

src/primitives/rectangles.jl
knuesel: What do you mean? You can check in the REPL that what I wrote is correct: `Rect2{T}` by itself is not valid Julia, but `Rect2 == (Rect{2,T} where T)` is exactly right (you can evaluate it, it returns `true`).

jkrumbiegel:
Valid points, I agree

src/primitives/rectangles.jl
putianyi889: Why is `Rect2{T}` not valid?
```julia
julia> Rect2{Float64}
Rect2{Float64} (alias for HyperRectangle{2, Float64})
```
While it's true, is there a Julia reference that `Type{A,B} == Type{A}{B}`? I wasn't able to find it and it apparently doesn't work on `Tuple`.

anyway, what I meant was that `RectT` is a standalone alias while you deleted it. I didn't realise that `RectT{T,N}` is basically `Rect{N,T}` and I've updated the docstring.

As for other changes, to me, it's more clear to phrase `Rect2{T} = Rect{2,T}` rather than `Rect2 = Rect{2,T} where T` as the former gives an example on how to use the alias. I'd like to hear opinions from others so I'll leave this unresolved.

jkrumbiegel:
Is there stuff exported from Makie where the suffix 2d is currently used to mean two-dimensional?

src/primitives/rectangles.jl
knuesel: Sorry I meant that `Rect2{T}` with literal `T` is not valid in itself, i.e. `T` has no special meaning in Julia, you have to define it first (or declare it as a type parameter using `where`). As you say `Rect2{Float64}` is perfectly valid, I think that would be a good example to use here rather than something with a generic `T`, because concrete examples are easier to grasp for beginners. I've edited my suggestion to add it.

Yes the `Type{A,B} == Type{A}{B}` rule is documented here: https://docs.julialang.org/en/v1/manual/types/#UnionAll-Types

> The type application syntax `A{B,C}` requires `A` to be a `UnionAll` type, and first substitutes `B` for the outermost type variable in `A` . The result is expected to be another `UnionAll` type, into which `C` is then substituted. So `A{B,C}` is equivalent to `A{B}{C}` .

I think tuples are a special case because `Tuple{Int}` is already a concrete type so the result doesn't expect an additional type parameter.

> As for other changes, to me, it's more clear to phrase Rect2{T} = Rect{2,T} rather than Rect2 = Rect{2,T} where T as the former gives an example on how to use the alias.

Note that `Rect2` can be used alone, for example `endcorner(r::Rect2) = origin(r) + widths(r)` is valid. But I see your point, I agree your version is better.

putianyi889: Then I'll close this conversation.

Off topic: turns out that `Tuple` is not `UnionAll`.
```julia
julia> Tuple <: UnionAll
false
```

knuesel: (please mark this as resolved, I don't see how to do it myself)

knuesel (commented):
Just one thing to clarify for this PR: do we want to document `RectT`, or is it rather meant as internal? It's not exported and I don't think I've seen it in user code so far.

SimonDanisch:
I think it's internal, but we could still mention it shortly?
But I wouldn't want to export it.

src/primitives/rectangles.jl
```diff
@@ -14,6 +14,19 @@ end
 ##
 # Constructors & typealiases
 
+"""
+    Rect{N,T} = RectT{T,N} = HyperRectangle{N,T}
```
knuesel, line 18:
```suggestion
    const Rect{N,T} = HyperRectangle{N,T}
```

src/primitives/rectangles.jl
```diff
@@ -14,6 +14,19 @@ end
...
+
+A rectangle in N dimensions, formally the cartesian product of intervals. See also [`HyperRectangle`](@ref). Its aliases are
+
+|        |`T`(eltype)|`Float32` |`Int`     |
+|--------|-----------|----------|----------|
+|`N`(dim)|`Rect{N,T}`|`Rectf{N}`|`Recti{N}`|
+|`2`     |`Rect2{T}` |`Rect2f`  |`Rect2i`  |
+|`3`     |`Rect3{T}` |`Rect3f`  |`Rect3i`  |
+"""
```
knuesel, line 27:
```suggestion

There is an additional unexported alias `RectT` that simply reverses the order of type parameters: `RectT{T,N} == Rect{N,T}`.

"""
```

knuesel:
I guess this doc PR can be merged and new aliases can be introduced in a new PR if desired?

SimonDanisch:
Yeah I was thinking about it and i was leaning towards merging it now :)

PR #209, opened 2024-01-22
Pass compat tests from Aqua

putianyi889:
[This bug](https://github.com/JuliaTesting/Aqua.jl/issues/75) is happening somehow


visr:
I think we are supposed to declare stdlib compatibilities now, see https://discourse.julialang.org/t/psa-compat-requirements-in-the-general-registry-are-changing/104958.

Project.toml
```diff
@@ -21,6 +21,12 @@ IterTools = "1.3.0"
 StaticArrays = "0.12, 1.0"
 StructArrays = "0.6"
 Tables = "0.2, 1"
+LinearAlgebra = ">=0.0.1"
+Test = ">=0.0.1"
+Random = ">=0.0.1"
+Aqua = ">=0.0.1"
+GeoJSON = ">=0.0.1"
+OffsetArrays = ">=0.0.1"
```
visr, line 29: Based on the Discourse link I sent stdlibs should get `<0.0.1,1`, not `>=0.0.1`. This applies to LinearAlgebra, Test and Random. The other 3 are not stdlibs and should probably be set to the latest breaking release.

It's also good practice to sort all the compat entries. Thanks!

visr (approved):

PR #212, opened 2024-02-05
Move the GeoInterface.jl code to an extension

asinghvi17:


SimonDanisch:
What's the goal for this? I think GeoInterface is a pretty small dependency ...

asinghvi17:
It was intended to remove a circular dependency but we don't actually need it.

The issue was that any other geometry in e.g. ArchGDAL, LibGEOS, etc. is auto-enabled for conversion into Makie, but GeoInterface's wrapper geometries are not (and cannot be) since the capability to do so lies in GeoInterfaceMakie.jl, which depends on GeoInterface.

What I can (probably) do is manually hook the GeoInterface geometries into the conversion system from Makie.

PR #213, opened 2024-02-08
add macro to generate aliases

putianyi889:
It's more like a proof of concept. Due to the discussion from #208 it's possible that the naming convention changes in the future and this PR tries to simplify that process.

### DIFF
--- a/src/GeometryBasics.jl
+++ b/src/GeometryBasics.jl
@@ -7,6 +7,8 @@ using EarCut_jll
 
 using Base: @propagate_inbounds
 
+include("utils.jl")
+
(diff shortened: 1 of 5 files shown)


SimonDanisch:
Ah, I was too slow, I also just took a stab at this...
I just decided against macros / eval, to finally have type aliases that are recognized by the vscode linter...

SimonDanisch:
https://github.com/JuliaGeometry/GeometryBasics.jl/pull/214

SimonDanisch:
Merged in https://github.com/JuliaGeometry/GeometryBasics.jl/pull/214

PR #214, opened 2024-02-08
dont use macro for aliases and introduce more variants (Float64 -> d, UInt -> ui)

SimonDanisch:
This PR makes all aliases recognizable by the linter and adds missing ones.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -106,3 +106,40 @@ Use `GeometryBasics.mesh` to get a mesh directly from a geometry:
 ```@repl quickstart
 mesh = GeometryBasics.mesh(rect)
 ```
+
+
+## Aliases
+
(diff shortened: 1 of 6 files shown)


SimonDanisch:
Weird, this makes the texture atlas hashing tests fail in Makie...
Not sure how, since this should only change aliases - and looks like all involved aliases seem to work correctly.
Will need some further investigation.

Meanwhile, we should think about adding even more aliases here for the other primitives?
Or also for e.g. Int32 etc? Although I think that goes a bit too far, but then again it would make it more consistent.

putianyi889:
As for the docstrings, it's now possible to generate one docstring for each alias. Then there should be a section on doc that clarifies the meaning of the suffixes.

PR #219, opened 2024-08-30
Geometry Basics refactor

ffreyer:
This is a continuation of #173.

I'm planning to keep the more verbose/drastic changes separate from #173 by doing them here so that we can easily do a full reset if need be. Maybe this also makes it easier to see what actually changed. For changes that essentially just fix #173 I'll try to update that pr.

# Plans

(Consider the code here pseudocode)

## Mesh type

My current plan for the `Mesh` type is roughly:

```julia
struct Mesh{D, T, Names..., ATs .<: AbstractVector}
    coordinates::AbstractArray{Point{D, T}}
    vertex_attributes::NamedTuple{Names, ATs}
    faces::Vector{FT}
    views::Vector{UnitRange}
end
```

- `coordinates` are always included so that the Mesh has a Dimension and element type
- `vertex_attributes` includes all other vertex data, e.g. normals, uvs, colors, etc. The current plan is for them to always be an `AbstractVector` and maybe `AbstractVector{<: Union{T, StaticVector{T}}` (I'll figure out the correct syntax for this later)
- `faces` (or connectivity) includes a vector of faces
- `views` includes UnitRanges that access sections of faces. This is for compatibility with meshes that define different sub meshes with e.g. different materials, i.e. for https://github.com/JuliaIO/MeshIO.jl/pull/95. We're also planning/thinking about adding views when merging meshes which would allow you to separate them again.

## Mesh wrapper / MetaMesh

As mentioned with `views` not all the data that you may want to bundle with a mesh is vertex data. You may want materials that apply per view, or maybe some mesh-global data like a name or reference to a file. For this we are thinking about adding a very general, simple wrapper:

```julia
struct MetaMesh{D, T, MeshT <: AbstractMesh{D, T}} 
    mesh::MeshT
    meta::Dict{Symbol, Any}
end
```

GeometryBasics will probably not touch this metadata at all. It's basically just there are storage which you can use for whatever you want. For obj loading in MeshIO this would be used to store materials, texture references, etc so that Makie can extract them and apply them when plotting a mesh.

## MultiFace and per-face Attributes

Wavefront obj files allow for faces to carry different indices for coordinates, normals and texture coordinates. We're planning to bring this to GeometryBasics via a `MultiFace` type:

```julia
struct MultiFace{FaceType, Names}
    face::FaceType
    attrib_faces::NamedTuple{Names..., FaceType}
end
```

- mirroring `Mesh`, the `face` would map to `coordinates` and always be there
- `attrib_faces` would map to the other `vertex_attributes` in a mesh
- We are using a NamedTuple here and in Mesh so we can verify that the faces and the vertex attributes match up. 

MeshIO already has Code that essentially transforms MultiFace to normal faces (i.e. addressing each vertex attribute with the same index). We are planning to move that here and simplify MeshIO a little by doing so. 

Adding MultiFace also means that we can use it for primitives which provides a solution for per-face Attributes while avoiding duplication. For example a Rect3 could define `coordinates()` ax 8 points, `normals()` as 6 vecs, and have faces defined as:
```julia
MultiFace{QuadFace}((1, 2, 3, 4), (1, 1, 1, 1)),
MultiFace{QuadFace}((1, 2, 5, 6), (2, 2, 2, 2)),
...
```
MeshIO's remapping code would then take care of the vertex duplication needed for rendering. This would be also be an optional conversion, i.e. a generic `mesh(Rect3(...))` would keep the 8 coordinates, 6 normals and 6 MultiFace's.

This may also be helpful for dealing with per-face normal generation and per-face attributes in general. E.g. we could have `decompose(FaceNormal(), mesh)` and some function that merges per-face attributes with a mesh by generating MultiFace's. (I haven't thought that much about this yet.)

## Type Tree

I don't have strong opinions on the type tree and types (beyond Mesh), but some things I'm considering are:
- restrict `AbstractFace` to Integer elements
- change `MultiType` from structs to alias `const MultiType = Vector{Type}`
- make `Simplex` a `GeometryPrimitve` like Rect, Sphere, etc. This would safe us reinventing the wheel with them, I think

<details>

<summary>Current Type Tree</summary>

Concrete types are bold, 

- AbstracGeometry{D, T}
	- GeometryPrimitive{D, T}
		- Rect, Sphere, etc
	- Polytope{D, T}
		- AbstractPolygon{D, T}
			- **Ngon{D, T, N}**
				- Line{D, T, 2}
				- Triangle{D, T, 3}
					- Triangle3d{3, T, 3}
						- GLTriangleElement{3, Float32, 3}
				- Quadrilateral{D, T, 4}
			- **Polygon{D, T}**
		- AbstractSimplex{D, T}
			- **Simplex{D, T, N}**
				- NSimplex{N}{D, T}
				- Tetrahedron{3, T, 4}
	- **MultiPolygon{D, T}**
	- **LineString{D, T}**
	- **MultiLineString{D, T}**
	- **MultiPoint{D, T}**
	- AbstractMesh{D, T}
		- **Mesh{D, T, VerticesT, FacesT}**
		- **MetaMesh{D, T, MeshT, Names, Types}**
- StaticVector{N, T}
	- **Point{N, T}** @fixed_vector
		- shorthands...
	- **Vec{N, T}** @fixed_vector
		- shorthands...
	- AbstractFace{N, T}
		- AbstractSimplexFace{N, T}
			- **SimplexFace{N, T}** @fixed_vector
		- AbstractNgonFace{N, T}
			- **NgonFace{N, T}** @fixed_vector
				- LineFace{2, T}
				- TriangleFace{3, T}
					- GLTriangleFace{3, GLIndex}
				- QuadFace{4, T}
- **Mat{Row, Col, T, L}** <: AbstractMatrix{T}

</details>

## Other changes

We also plan to:
- move Ray from Makie
- move Plane from Makie
- consistently define coordinates or 2D primitives in counter-clockwise order for Polygon construction
- make sure simplexes and other things for FEM exist and work correctly (probably not adding anything new here though, we can do that later)


SimonDanisch:
Yes, we have discussed it a lot over the last year and looked at how to use use it for Makie as the main Geometry package multiple times.
While it's a great package that offers a lot of algorithms, it doesn't offer some pretty basic things we need to have for basic graphics programming (materials, material groups, uvs, multifaces etc). 
The Mesh constructor documentation is also pretty short and doesn't really show how to construct a mesh with uvs and normals - which is the main thing we care about.

It's also very opiniated for some things, that we don't really care about which will break 90% of Makie's existing code (e.g. forcing Points to have a unit).

I think Meshes is more of a computational Geometry package with focus on correctness, while GeometryBasics is really more for basic computer graphic types, where you really just care about loading an OBJ file and uploading it to the GPU correctly and efficiently.
Anything beyond those basic needs seem like a time intensive distraction, from a Makie perspective.

We also have a few, admittedly odd things in GeometryBasics, which reach quite deeply into Makie, and I'm sure you wouldn't want to carry those over to Meshes.jl - and I'm pretty sure I dont want to refactor them right now in Makie.

The fact, that there is usually zero room for compromises in that regard with Meshes.jl makes it also pretty scary to rely on it, in case we need something fixed quickly in a non perfect way.

I think all the design decisions in Meshes are thoughtful and make a lot of sense for what you're doing, but it doesn't really work out as well for OpenGL programming.

Believe me I thought long and hard about it, but in the end it seemed easier and less work to just improve GeometryBasics a bit, and simply convert from Meshes.jl to GeometryBasics for any tasks that require more intense computational operations on meshes.

If you think it's less work and more correct to replace GeometryBasics.jl with Meshes.jl in Makie, we're happy to review a PR to Makie, which replaces all functionality with Meshes.jl and keeps all of Makie's functionality intact ;)

juliohm:
> While it's a great package that offers a lot of algorithms, it doesn't offer some pretty basic things we need to have for basic graphics programming (materials, material groups, uvs, multifaces etc).

Thanks for sharing requirements. I understand that there are constraints on where to store the material/uv information for the underlying GL calls.

Meshes.jl currently assumes that all metadata is stored outside the mesh as Tables.jl tables, including vertex and face tables. This is decoupled nowadays into a separate package called GeoTables.jl. I believe that you could achieve something similar by creating a wrapper type MakieMesh that stores a Meshes.Mesh + uv + material. That way you could benefit from geometry+topology algorithms that are thoroughly tested, and spend time/energy on the material + uv related to graphics and visualization:

```julia
using Meshes: Mesh

struct MakieMesh{UV,Material}
  mesh::Mesh
  uv::UV
  material::Material.
end
```

> It's also very opiniated for some things, that we don't really care about which will break 90% of Makie's existing code (e.g. forcing Points to have a unit).

That is fair. We need units in our vertex coordinates to make algorithms more robust. They are always there, it is just that we make them explicit with Unitful.jl so that users can control objects sizes (in, cm, m, ...) more easily. If you end up using Meshes.jl algorithms, you rarely need to specify units by hand. Units are mostly present in user-input.

> I think Meshes is more of a computational Geometry package with focus on correctness, while GeometryBasics is really more for basic computer graphic types, where you really just care about loading an OBJ file and uploading it to the GPU correctly and efficiently.
> Anything beyond those basic needs seem like a time intensive distraction, from a Makie perspective.

Makes sense. I am just trying to see if you could leverage what we built. It is a lot of work to write and test these algorithms, so I am trying to save you guys from going down that time-sinking route.

>We also have a few, admittedly odd things in GeometryBasics, which reach quite deeply into Makie, and I'm sure you wouldn't want to carry those over to Meshes.jl - and I'm pretty sure I dont want to refactor them right now in Makie.

Could this be addressed with the wrapper type above maybe?
  
> The fact, that there is usually zero room for compromises in that regard with Meshes.jl makes it also pretty scary to rely on it, in case we need something fixed quickly in a non perfect way.

Totally understand it. Our pedantic review has two sides to it. On one side you get robust code, and a design that is carefully shaped for general use besides visualization. On the other side I can see how this inhibits quick (1 day) fixes.

> I think all the design decisions in Meshes are thoughtful and make a lot of sense for what you're doing, but it doesn't really work out as well for OpenGL programming.

If you can list what is limiting in the current design of Meshes.jl for OpenGL programming, we can try to find a solution. I believe that a wrapper type can go a long way.

> If you think it's less work and more correct to replace GeometryBasics.jl with Meshes.jl in Makie, we're happy to review a PR to Makie, which replaces all functionality with Meshes.jl and keeps all of Makie's functionality intact ;)

I wish I could find some time to sit down and work on this, but it is hard to adjust Makie.jl code without deep knowledge of its internal design. It would be much easier for a Makie dev to try this experiment.

termi-official:
> Have you considered migrating to Meshes.jl instead? 

Yes (from Ferrite perspective) and I really appreciate the effort put into this package. However, we figured out that we just need an abstract mesh interface to simplify integration with visualizers downstream. This probably also holds true for all other PDE packages. With this in mind Meshes.jl can be interesting in the sense that we could use it to generate meshes for e.g. finite element simulations, but there are no algorithms for high quality meshes included (where I mean with quality as in finite element quality measures).

Furthermore, I should note that the biggest pain on our end, regarding simulation, is the efficient management of nonlinearities (e.g. quadratic hexahedra as elements or higher order solution fields).

juliohm:
> Yes (from Ferrite perspective) and I really appreciate the effort put into this package. 

Thank you. We develop Meshes.jl as part of our jobs, and plan to keep working actively on it for a very long time. 

> However, we figured out that we just need an abstract mesh interface to simplify integration with visualizers downstream.

I have to say that I am afraid of interface efforts when not a single implementation is thoroughly tested. I've seen this happen multiple times in Julia: people spending years of development on interfaces that don't accommodate all the features needed by downstream applications.

At some point Makie needs a concrete implementation. My suggestion above is to try to build a wrapper type on top of Meshes.jl that meets all Makie.jl requirements. If that is not possible, that is fine. We will have evidence that the designs are indeed incompatible.

When a feature is missing for a specific downstream application, there are two possible routes:

1. contribute the feature directly to the concrete implementation that has been chosen as the base
2. design an abstract interface that will be partially supported by a subset of implementations

In my opinion, the option (2) is rarely the best option. It leads to incomplete, poorly covered implementations. The Julia community is tiny, so option (1) has a higher chance of positive outcomes.

ffreyer:
> > I think all the design decisions in Meshes are thoughtful and make a lot of sense for what you're doing, but it doesn't really work out as well for OpenGL programming.
> 
> If you can list what is limiting in the current design of Meshes.jl for OpenGL programming, we can try to find a solution. I believe that a wrapper type can go a long way.
>
> > If you think it's less work and more correct to replace GeometryBasics.jl with Meshes.jl in Makie, we're happy to review a PR to Makie, which replaces all functionality with Meshes.jl and keeps all of Makie's functionality intact ;)
> 
> I wish I could find some time to sit down and work on this, but it is hard to adjust Makie.jl code without deep knowledge of its internal design. It would be much easier for a Makie dev to try this experiment.

Personally I'm rather intimidated by Meshes.jl too. I did not have any lectures on geometry in university so a lot of the nomenclature is new to me and that can make the docs hard to follow at times. There are also a lot of files and folders that I don't know my way around.

I think trying Meshes.jl in Makie will also come with a lot of complications. Treating Points and Vecs more accurately will certainly generate a bunch of errors. Always having units and coordinate systems with points may break `convert_arguments`, our unit system and Float32 converts. Maybe it also causes issues with OpenGL or Cairo though I don't think it would. The lack of `Vec{N, Int}` will break code like `matrix[Vec(1,2,3), Vec(1,2,3)]`. In general I think the always spatial (?) Points from Meshes.jl just won't work well with Makie, since there are a lot more units that make sense in a plot.

With respect to meshes it sounds like we'd have to create new mesh types/wrappers that include extra vertex attributes which would need to somehow integrate with your algorithms, and then also port a bunch of the utilities from here. E.g. generating normals from points and faces, and generating normals and uvs from primitives for which would also need to integrate with your discretization code somehow. I'd probably want most of this pr to get ported as well, since `MultiFace` is my idea of dealing with per-face normals (among other things) and the `MetaMesh` changes are in part there to enable loading and displaying more complex models in Makie. 

To me it seems like we'd just end up rewriting half of GeometryBasics on top of Meshes. I'd rather adjust GeometryBasics to be a common ground between Makie and Meshes then. Though I feel like that would also be difficult.

SimonDanisch:
I do think the conversion works pretty well, which lowers the pressure...

juliohm:
> I did not have any lectures on geometry in university so a lot of the nomenclature is new to me and that can make the docs hard to follow at times.

If you have specific questions in mind, we can try to help. The high-level interface is usually very easy to use.

> Treating Points and Vecs more accurately will certainly generate a bunch of errors.

These may be good errors. Better catch them early.

> Always having units and coordinate systems with points may break convert_arguments, our unit system and Float32 converts.

I am not following the internal details closely, but all our recipes work well with Makie.jl. Whenever the built-in recipe doesn't support units, we call `ustrip` to strip the units. Also, our tests are run with both Float32 and Float64.

> The lack of Vec{N, Int} will break code like matrix[Vec(1,2,3), Vec(1,2,3)].

Break in what sense? You can construct a `Meshes.Vec(1,2,3)` from Int literals, they will be converted to float. If you need something else that is not a geometric vector, but simply a static container of numbers (e.g. StaticArray), then you need a another type. `Meshes.Vec` is a geometric vector.

> With respect to meshes it sounds like we'd have to create new mesh types/wrappers that include extra vertex attributes which would need to somehow integrate with your algorithms

Yes, that would be a direction worth exploring.

> and then also port a bunch of the utilities from here. E.g. generating normals from points and faces, and generating normals and uvs from primitives for which would also need to integrate with your discretization code somehow

We provide `normal(triangle)` and other geometries already. Same for `discretization`, it should be plug and play if you call these methods on Meshes.Geometry or Meshes.Mesh.

Let me know if you need any help with the functionality in Meshes.jl. Hope you can find a good direction that saves you time and meets Makie.jl requirements. We love visualizing our Meshes.jl objects with Makie.jl

PR #225, opened 2024-10-30
Add 3D capability to earcut_triangulate

asinghvi17:
This method assumes three things:
- The first three points in the polygon are not NaN
- The first three points in the polygon are all different
- The polygon lies in a single plane, that can be described by its first three points

It simply creates a plane from the first three points of the polygon, and projects the rest of the polygon onto it.  Then, EarCut can triangulate the polygon by the usual method for 2D.

This will enable plotting 3D polygons in GLMakie and WGLMakie with no extra steps, CairoMakie needs some more fixes it seems.  But we're getting there!

Fixes https://github.com/MakieOrg/Makie.jl/issues/4015, and the other issue someone raised about a set of polygons in 3D that I can't find anymore.

Probably needs tests, but not sure what a good test suite would be?  Ideas appreciated :)

cc @milankl - this is the algorithm I sent you on Slack

### DIFF
--- a/src/triangulation.jl
+++ b/src/triangulation.jl
@@ -208,6 +208,33 @@ function earcut_triangulate(polygon::Vector{Vector{Point{2,Int32}}})
     return unsafe_wrap(Vector{GLTriangleFace}, array[1], array[2]; own=true)
 end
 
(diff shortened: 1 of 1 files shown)


asinghvi17:
This might be better placed in Makie - closing for now.

PR #226, opened 2024-10-30
Bump version to 0.5 (breaking version)

asinghvi17:


asinghvi17:
Sure, but that would probably merit a patch release from before the refactor with a deprecation warning as well?

t-bltg:
@asinghvi17, done in https://github.com/JuliaGeometry/GeometryBasics.jl/pull/227/commits/34f40a25cd4d67c7a6e35228fd41a0894695229b.

ffreyer:
I talked with Simon and we think it's better to just do the Tesselation -> Tessellation change with the breaking release.

ffreyer (approved):

ffreyer:
In case people come here for some notes on what changed/how to update:

**Changes that could break things**
- `Tesselation` -> `Tessellation`
- `meta()` removed: `Mesh(meta(kwargs...), faces)` -> `Mesh(faces; kwargs...)`
- `metafree()` removed: no replacement needed
- `mesh.normals` -> `mesh.normal`
- `mesh(..., normals = ...)` -> `mesh(..., normal = ...)`
- `Mesh` type name/structure has been simplified. See `basic_types.jl` or just check by hand
- `Rect2`  now generates coordinates in counter-clockwise order
- `Cylinder` , `Rect3`, `Pyramid` now generate `FaceView`'s. You can remove these with `expand_faceviews(mesh)`. Some consequences:
	- `coordinates(::Rect3)` are no longer duplicated
	- `normals(::Rect3)` and `texturecoordinates(::Rect3)` are FaceViews to avoid duplication. If you need them synchronized with coordinates use `normals(expand_faceviews(uv_normal_mesh(Rect3(...))))`
- MeshIO.jl may output a `MetaMesh` for `obj` files now. This includes additional non-vertex data (e.g. materials). The "raw" mesh is in `metamesh.mesh`.
- MeshIO.jl may generate a `Mesh` with `FaceViews`

**Other changes**
- `Cylinder` no longer renders with rounded caps (due to no longer sharing normals between mantle and caps)
- `merge(meshes)` now also sets `mesh.views` which acts as a collection of index ranges into `mesh.faces` corresponding to the input meshes. `split_mesh()` can now revert a merge.
- `mesh(input_mesh; pointtype, facetype, vertex_attributes...)` can be used to change the pointtype or facetype, or add additional vertex attributes to a given mesh. (This will create a new mesh)
- added `face_normals(positions, faces)` for generating per-face normals (uses `FaceView`)

Issue #224, opened 2024-10-28
Rename `Tesselation` -> `Tessellation`


t-bltg:
The correct spelling is `Tessellation`, requires a breaking change.

Per https://github.com/MakieOrg/Makie.jl/pull/4522#issue-2611882857.

PR #227, opened 2024-10-30
rename `tesselation` to `tessellation`

t-bltg:
Fix https://github.com/JuliaGeometry/GeometryBasics.jl/issues/224.

EDIT: now non-breaking: the deprecated binding `Tesselation` is kept and can be removed later.

### DIFF
--- a/src/interfaces.jl
+++ b/src/interfaces.jl
@@ -58,58 +58,58 @@ something like `Vec2f`.
 texturecoordinates(primitive, nvertices=nothing) = nothing
 
 """
-    Tesselation(primitive, nvertices)
+    Tessellation(primitive, nvertices)
 
 When generating a mesh from an abstract geometry, we can typically generate it
 at different levels of detail, i.e. with different amounts of vertices. The 
-`Tesselation` wrapper allows you to specify this level of detail. When generating
+`Tessellation` wrapper allows you to specify this level of detail. When generating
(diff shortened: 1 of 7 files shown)


src/interfaces.jl
```diff
@@ -58,58 +58,60 @@ something like `Vec2f`.
...
 ```
 """
-struct Tesselation{Dim,T,Primitive,NGrid} <: AbstractGeometry{Dim, T}
+struct Tessellation{Dim,T,Primitive,NGrid} <: AbstractGeometry{Dim, T}
     primitive::Primitive
     nvertices::NTuple{NGrid,Int}
 end
 
-function Tesselation(primitive::GeometryPrimitive{Dim,T},
+Base.@deprecate_binding Tesselation Tessellation
```
EdsterG, line 88: Makie has recently refactored away from this form of deprecation to fix some precompile warnings. I'm not deeply familiar with the root cause and just pointing it out incase it also applies here. See https://github.com/MakieOrg/Makie.jl/pull/4333.

t-bltg: @SimonDanisch or @jkrumbiegel, what should we do here ?

ffreyer: I think that was a problem because of the re-exporting Makie backends do. That causes these deprecations to trigger unnecessarily. I don't think it's an issue here.

ffreyer (approved):

Issue #207, opened 2023-12-28
GeometryBasics.Mesh incorrectly converts quads to triangles


Kevin-Mattheus-Moerman:
When trying to convert quadrilateral faces `F` and their vertices `V` to a mesh using:
```julia
M=GeometryBasics.Mesh(V,F)
``` 
One correctly obtains a quadrilateral mesh description: 
```julia
julia> typeof(GeometryBasics.Mesh(V,F))
GeometryBasics.Mesh{3, Float64, GeometryBasics.Ngon{3, Float64, 4, Point3{Float64}}, SimpleFaceView{3, Float64, 4, Int64, Point3{Float64}, QuadFace{Int64}}}
```

However an error occurs if the mesh only consists of a single quad face, i.e. `length(F)==1`, in this case it looks like that quad is instead interpreter as a triangle: 

```
julia> typeof(GeometryBasics.Mesh(V,F[1]))
GeometryBasics.Mesh{3, Float64, GeometryBasics.Ngon{3, Float64, 3, Point3{Float64}}, FaceView{GeometryBasics.Ngon{3, Float64, 3, Point3{Float64}}, Point3{Float64}, TriangleFace{Int64}, Vector{Point3{Float64}}, Base.ReinterpretArray{TriangleFace{Int64}, 1, Tuple{Int64, Int64, Int64}, TupleView{Tuple{Int64, Int64, Int64}, 3, 1, QuadFace{Int64}}, false}}}
```

A longer example and some context is given below.  

Here is an example quadrilateral mesh description, in this case for a cube. 

```julia 
    # Create vertices       
    V=Vector{GeometryBasics.Point{3, Float64}}(undef,8)
    s = r/sqrt(3.0)
    V[1 ] = GeometryBasics.Point{3, Float64}( -s, -s, -s)
    V[2 ] = GeometryBasics.Point{3, Float64}( -s,  s, -s)
    V[3 ] = GeometryBasics.Point{3, Float64}(  s,  s, -s)
    V[4 ] = GeometryBasics.Point{3, Float64}(  s, -s, -s)
    V[5 ] = GeometryBasics.Point{3, Float64}( -s, -s,  s)
    V[6 ] = GeometryBasics.Point{3, Float64}( -s,  s,  s)
    V[7 ] = GeometryBasics.Point{3, Float64}(  s,  s,  s)
    V[8 ] = GeometryBasics.Point{3, Float64}(  s, -s,  s)
        
    # Create faces
[...]

Issue #155, opened 2021-12-05
Intersections of lines with Int parameters crash


racinmat:
```julia
using GeometryBasics
intersects(Line(Point(-1, 0), Point(1, 0)), Line(Point(0, -1), Point(0, 1)))
```
crashes with
```
julia> intersects(Line(Point(-1, 0), Point(1, 0)), Line(Point(0, -1), Point(0, 
1)))
ERROR: LoadError: MethodError: no method matching Int64(::Irrational{:π})      
Closest candidates are:
  (::Type{T})(::T) where T<:Number at C:\Users\racinsky\AppData\Local\Programs\Julia-1.7.0\share\julia\base\boot.jl:770
  (::Type{T})(::AbstractChar) where T<:Union{Int32, Int64} at C:\Users\racinsky\AppData\Local\Programs\Julia-1.7.0\share\julia\base\char.jl:51
  (::Type{T})(::AbstractChar) where T<:Union{AbstractChar, Number} at C:\Users\racinsky\AppData\Local\Programs\Julia-1.7.0\share\julia\base\char.jl:50        
  ...
```

PR #238, opened 2025-02-03
Improve test coverage

ffreyer:
This pr aims to increase test coverage and fix problems I hit along the way. WIP for now.

Questions:
- Should `Rect()` default to `Rect(NaN..., 0...)` instead of `Rect(Inf..., -Inf...)`? Otherwise, how should utility functions treat infinite Rects? E.g. should volume return NaN? minimum and maximum? Should we consider `-Inf` widths as invalid?
- Does it make sense for `mat[Rect2i(1, 1, 2, 2)]` to index `[2:3, 2:3]`, i.e. be zero based?

(Significant) (Code) Changes:
- Rect:
    - refactored constructors to generalize them and make them depend more on each other. (This should only add more constructors)
    - generalized `euclidean_distance`, `update` to accept `VecTypes`
- added `strictly_equal_face_vertices()` for comparing the result of `split_mesh` with pre-merge meshes. This considers two meshes to be equal if `a.vertex_data[a.face] == b.vertex_data[b.face]` holds for every face and every vertex attribute.
- removed `merge_vertex_indices(faces...[, index_counter])` method which I added as an alternative to `merge_vertex_indices(tuple_of_face_arrays[, index_counter])` but never used. This function is internal with the (more) public interface being `expand_faceviews()`, so this should not be breaking. (It's also not used in MeshIO or Makie)
- **breaking** Removed `Rect{T}(obj)` constructor for boundignboxes as it clashes with the `Rect{N, T}` definition of the type. The implementations now use `Rect{N, T}(obj)` with other methods funneling into that one. This enables `Rect{N}(obj)` and `RectT{T}(obj)` analogous to other Rect constructors.
- ~~**maybe breaking**~~ reworked `self_intersections()` to allow a segment to intersect with multiple other segments. ~~For this, the format of the returned indices changed from `Vector{Int}` to `Vecto{Tuple{Int, Int}}`.~~ This is now an internal change. Also did some low hanging performance optimizations (skip redundant half of the loop)

Test TODO;
- [x] Rect (file: 57% -> 96%, total: 64% -> 71%)
    - [x] constructors, convert, centered, split, bbox-like constructors
    - [x] euclidean distances
    - [x] Rect as matrix index
    - [x] Matrix - Rect multiplications
- [x] bits and pieces in basic_types.jl (file: 63% -> 83%, total: 74%)
- [x] most of triangulation.jl (file: 26% -> 90%, total: 80%)
- [x] meshes.jl (file: 68% -> 85%, total: 82%)
    - [x] split_mesh
    - [x] duplicate face removal
    - [x] map_coordinates
- [x] boundingboxes.jl (file: 48% -> 88%, total: 83%)
- [x] lines.jl (file: 34% -> 98%, total: 86%)
- [x] spheres.jl (file: 63% -> 100%, total: 86%) 
- [x] OffsetIntegers (file: 61% -> 55% ???) and FixedArrays (unchanged) (total: 86%)


asinghvi17:
Do we actually have codecov enabled on this repo?  Should we do that, if the coverage is now worth showing off?  :D

ffreyer:
We do, it just doesn't run in prs

ffreyer:
Well it does, it just errors

src/boundingboxes.jl
```diff
@@ -1,18 +1,41 @@
...
+# Implementations
+# Specialize fully typed Rect constructors
+function Rect{N, T}(geom::AbstractGeometry) where {N, T <: Number}
+    if applicable(Rect{T}, geom)
+        @warn "`Rect{T}(geom)` is deprecated as the final boundingbox method. Define `Rect{N, T}(geom)` instead."
+        return Rect{T}(geom)
+    else
+        return Rect{N, T}(coordinates(geom))
+    end
+end
```
ffreyer, line 20: I don't think there is much of a point to making a user facing depwarn here, because on master `Rect{T}(geom)` only works if T matches the input eltype. I therefore doubt anyone relied on it. It also throws an ArgumentError lower down the line (and would create a stackoverflow with this dep warning.)

For devs I think it makes more sense, because I think it's fairly likely someone would copy the syntax from GeometryBasics. And without a warning, their implementation would silently be ignored.

ffreyer: nvm, doesn't work. I thought for a second that `N <: Val` was working for catching literals but I don't think it actually does. Without type specifity on `Rect{N}(...)` this will always call the applicable branch and loop.

ffreyer:
Fixes #155 (if it's not already fixed on master)

ffreyer:
Fixes #207

SimonDanisch (approved):

PR #245, opened 2025-03-07
Use shoelace approach to compute normals

Kevin-Mattheus-Moerman:
@ffreyer @SimonDanisch This PR implements the "shoelace" method for normal computation: 
image

The approach should reduce to the same thing for triangles but generalises properly for faces with more than 3 vertices. Previously 3 points were taken along the face to compute the normal. This assumes that the face/polygon is planar and that the chosen points are not for co-linear edges. However both can naturally occur for faces with more than 3 nodes. Previously such faces would have inaccurate normals computed or NaNs would be returned. 

Here is a test case for a Quad with a co-linear edge, the above implemented method now returns the z-dir as the normal direction properly: 
```
v = [Point{3,Float64}(0.0,0.0,0.0),Point{3,Float64}(1.0,0.0,0.0),Point{3,Float64}(2.0,0.0,0.0),Point{3,Float64}(2.0,1.0,0.0)]
f = [QuadFace{Int}(1,2,3,4)]
n = normals(v,f)
```
I have implemented other cases for testing too. Note that during testing for me this errored: 
``` 
    @test all(n -> n == values(ns)[end], values(f_ns)[16:end])
```
It seemed correct to withing 1e-8, So I changed it to use `≈`
```
    @test all(n -> n ≈ values(ns)[end], values(f_ns)[16:end])
```
Check if that is indeed what we'd like. 

Let me know if you are happy with this or if you need me to work on anything.

### DIFF
--- a/src/fixed_arrays.jl
+++ b/src/fixed_arrays.jl
@@ -145,6 +145,13 @@ Base.isnan(p::Union{AbstractPoint,Vec}) = any(isnan, p)
 Base.isinf(p::Union{AbstractPoint,Vec}) = any(isinf, p)
 Base.isfinite(p::Union{AbstractPoint,Vec}) = all(isfinite, p)
 
+function to_ndim(T::Type{<: VecTypes{N, ET}}, vec::VecTypes{N2}, fillval) where {N,ET,N2}
+    T(ntuple(Val(N)) do i
+        i > N2 && return ET(fillval)
+        @inbounds return vec[i]
+    end)
+end
+
 ## Generate aliases
 ## As a text file instead of eval/macro, to not confuse code linter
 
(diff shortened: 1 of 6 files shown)


Kevin-Mattheus-Moerman:
@ffreyer how's this coming along? Let me know if I can help.

ffreyer:
I think this is good to go. It would be good if you checked again if everything is still working. Maybe also add some basic tests for the things codecov is flagging as missing (see files changed)

Just fyi, I changed things around so that orthogonal_vector always returns a Vec3 now.

Kevin-Mattheus-Moerman:
> I think this is good to go. It would be good if you checked again if everything is still working. Maybe also add some basic tests for the things codecov is flagging as missing (see files changed)

Sure I'll have a look. 

> Just fyi, I changed things around so that orthogonal_vector always returns a Vec3 now.

Great, that makes sense.

PR #246, opened 2025-03-07
Use default getindex for `vector[face]` when no Polytope is generated

ffreyer:
This allows things like `f[f .== 1]`

### DIFF
--- a/src/basic_types.jl
+++ b/src/basic_types.jl
@@ -31,10 +31,6 @@ abstract type AbstractNgonFace{N,T} <: AbstractFace{N,T} end
 
 abstract type AbstractSimplex{Dim,T} <: Polytope{Dim,T} end
 
-@propagate_inbounds function Base.getindex(points::AbstractVector{P}, face::F) where {P<: Point, F <: AbstractFace}
-    return Polytope(P, F)(map(i-> points[i], face.data))
-end
-
 @fixed_vector SimplexFace = AbstractSimplexFace
 
 const TetrahedronFace{T} = SimplexFace{4,T}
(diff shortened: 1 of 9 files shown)


ffreyer:
The only thing that changes is that a tuple becomes a Vector

SimonDanisch:
Ok, I think this is the right approach to fix #249  and #247 
We also need to remove:
```julia
@propagate_inbounds function Base.getindex(points::AbstractVector{P}, face::F) where {P<: Point, F <: AbstractFace}
    return Polytope(P, F)(map(i-> points[i], face.data))
end

@propagate_inbounds function Base.getindex(elements::AbstractVector, face::F) where {F <: AbstractFace}
    return map(i-> elements[i], face.data)
end
```
Looks like that was the behaviour before, and I dont see a good reason why whe should deviate from the previous behavior.

Kevin-Mattheus-Moerman:
@ffreyer yep this seems to fix those issues, thanks! 

* #247 is fully fixed. 
* The core issue in #249 i.e. allowing for `f[f.!=3]` on faces, is fixed. 
 A side note is that if we have `f = QuadFace{Int64}(1,2,3,4)` then `f.!=3` still returns the odd creature that is a `QuadFace{Bool}`. Similarly for a Point{3,Float} by the name of `v` this `v.<0.5` returns a `Point3{Bool}`. Both these face/point types seem to never actually represent a valid face or point in any way. For my projects this is not breaking anything but it does seem like perhaps these things should return some kind of static vector containing Bools? Again this doesn't break anything, and perhaps you have good reasons for keeping it.

PR #257, opened 2025-04-16
Add Cone primitive

ffreyer:
I want to remove the mesh generation code in Makie that's used for arrows and add replacements as GeometryPrimitives here. This is coupled with https://github.com/MakieOrg/Makie.jl/pull/4925.

Currently this just adds `Cone()` with the idea of simplifying the arrow components to full, closed meshes.

### DIFF
--- a/src/primitives/Cone.jl
+++ b/src/primitives/Cone.jl
@@ -79,7 +79,19 @@ function normals(c::Cone, nvertices = 30)
     # cap
     ns[end] = Vec3f(normalize(c.origin - c.tip))
 
-    return ns
+    faces = Vector{GLTriangleFace}(undef, nvertices)
+
+    # shell
+    for i in 1:nhalf
+        faces[i] = GLTriangleFace(i, mod1(i+1, nhalf), nhalf+1)
+    end
+
+    # cap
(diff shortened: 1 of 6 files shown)


src/primitives/Cone.jl
```diff
@@ -0,0 +1,102 @@
...
+    norm = 1.0 / sqrt(1 + z*z)
+    for i in 1:nhalf
+        phi = (i-1) * step
+        ns[i] = R * (norm * Vec3f(cos(phi), sin(phi), z))
+    end
+
+    # tip - this is undefined / should be all ring angles at once
+    # for rendering it is useful to define this as Vec3f(0), because tip normal
+    # has no useful value to contribute to the interpolated fragment normal
+    ns[end-1] = Vec3f(0)
```
ffreyer, line 77: This fixes seams on the mesh. The left mesh uses triangles where the tip vertex is duplicated for each triangle with a different normal (average of bottom normals). The right uses one (0,0,0) vertex at the tip:
image
You get the same seams with quads instead of triangles and even if you give the tip a finite radius, i.e. with a conical frustum. (?)

SimonDanisch (approved):

PR #260, opened 2025-05-26
Breaking: remove geointerface

rafaqz:
This PR can be rolled with the next breaking release.

The reason is to break the circular dependencies between GeoInterface and GeometryBasics extensions in regards to plot recipes.

We will move all of this code to an extension in GeoInterface.jl. GeoInterfaceMakie.jl limits GeometryBasics.jl to 0.5 in most setups.  There is a (very) small chance to get both GeoInterface and GeometryBasics 0.6 at the same time (if basically no other JuliaGeo deps are installed), which may break some code. So we will need to have the GeoInterface version with GeometryBasics.jl 0.6 ready to tag straight away whenever this is tagged.

@evetion @SimonDanisch @asinghvi17

PR #263, opened 2025-07-05
Move GeoInterface.jl to an extension

rafaqz:
Revives #212 but for more recent changes here

Removing the GeoInterface dependency is actually crucial to simplifying GeoInterface plotting dependencies and Makie 0.24, because currently GeoInterface cant extend anything that depends on GeometryBasics.jl because of the circularity (i.e. it will break precompilation).

(there is some chance there will be circularity between extensions somewhere anyway after this, but lets deal with that later, and avoid the more drastic #260 option for now)

@SimonDanisch this is the start of getting rid of GeoInterfaceMakie.jl ;)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -31,7 +31,7 @@ PrecompileTools = "1.0"
 Random = "<0.0.1,1"
 StaticArrays = "0.6, 1"
 Test = "<0.0.1,1"
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"


rafaqz:
I dropped Julia 1.6 and set the minimum compat to 1.10, so that extensions work

rafaqz:
The documenter issue is just xvfb, not this PR: 
```
  DISPLAY=:0 xvfb-run -s '-screen 0 1024x768x24' julia --project=docs docs/make.jl
  shell: /usr/bin/bash -e {0}
  env:
    JULIA_PKG_SERVER: 
    GITHUB_TOKEN: ***
    DOCUMENTER_KEY: 
xvfb-run: error: Xvfb failed to start
Error: Process completed with exit code 1.
```

asinghvi17 (approved):

asinghvi17:
I think registrator should be fine?  At least I have never had that experience.

This isn't actually breaking, all the functionality still exists, it's just in an extension now.  The only potentially breaking thing might be people using `GeometryBasics.GeoInterface` but I don't imagine many people were doing that...

PR #264, opened 2025-07-26
GeoInterface improvements

asinghvi17:
## Summary

- Adds `GeoInterface.convert` for `LineTrait` → `GeometryBasics.Line` (2D and 3D)
- Extends `GeoInterface.convert` for `LineString` to also accept `LinearRingTrait`, so converting any `Polygon` (which has `LinearRing` exterior/holes) works correctly
- Adds `geointerface_geomtype(::LinearRingTrait) = LineString`
- Adds tests covering all new code paths (2D and 3D `Line` conversion, 2D and 3D `LinearRing` → `LineString` conversion)

### DIFF
--- a/ext/GeometryBasicsGeoInterfaceExt.jl
+++ b/ext/GeometryBasicsGeoInterfaceExt.jl
@@ -34,6 +34,7 @@ geointerface_geomtype(::GeoInterface.PointTrait) = Point
 geointerface_geomtype(::GeoInterface.MultiPointTrait) = MultiPoint
 geointerface_geomtype(::GeoInterface.LineTrait) = Line
 geointerface_geomtype(::GeoInterface.LineStringTrait) = LineString
+geointerface_geomtype(::GeoInterface.LinearRingTrait) = LineString
 geointerface_geomtype(::GeoInterface.MultiLineStringTrait) = MultiLineString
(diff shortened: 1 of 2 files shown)


ext/GeometryBasicsGeoInterfaceExt.jl
```diff
@@ -34,6 +34,7 @@ geointerface_geomtype(::GeoInterface.PointTrait) = Point
 geointerface_geomtype(::GeoInterface.MultiPointTrait) = MultiPoint
 geointerface_geomtype(::GeoInterface.LineTrait) = Line
 geointerface_geomtype(::GeoInterface.LineStringTrait) = LineString
+geointerface_geomtype(::GeoInterface.LinearRingTrait) = LineString
```
asinghvi17, line 37: Lol I just ran into this again, we should get this in!

rafaqz: I feel like I have also fixed this multiple times locally 😭

rafaqz: Get the codecov bot onto it? I'm not a member here

ffreyer:
This only affects GeoInterface so I don't think this needs review from the GeometryBasics side?

PR #276, opened 2026-01-31
Clean up Rect set operations (union, intersect, diff, overlaps)

ffreyer:
This is something I will probably need for optimizing an SDF `volume` plots in Makie. Doesn't need to be merged/tagged until I get that near completion

Changes:
- added `volumetric` kwarg to `isempty` to switch between "any" and "all" widths <= 0 checks
- exclude all-empty rects from union (probably to avoid `Rect()` including Inf in unions)
- add `bbox_diff`to calculate the bounding box of the pieces left after cutting one rect out of another (without explicitly calculating the pieces)
- return empty rect in `intersect` if rects have no overlap (+ type promotion)
- all one rect being fully inside the other in `overlaps`
- move rect tests into their own file, categorize tests, add tests for 
    - add tests for `bbox_diff`
    - add tests for `isempty`
    - extend tests for `in` (using explicit rects to go through specific situations rather than results from `splt`)
    - extend tests for `overlaps` to cover edge cases
    - add numeric tests for `union` (not just `isa Rect`)
    - add numeric tests for `intersect` (not just `isa Rect`)
    - add numeric tests for `update` (not just `isa Rect`)

TODO:
- consider removing Allen's Interval Algebra functions for general Rects, see below


ffreyer:
https://en.wikipedia.org/wiki/Allen%27s_interval_algebra
https://github.com/JuliaGeometry/GeometryBasics.jl/blob/9918b1881d059e16354ce44f0d66eaaa82fc5ae9/src/primitives/rectangles.jl#L441-L486

This stuff makes sense for intervals but I fail to see how this should generalize to rectangles. For examples saying one interval is `before` another is clear, but for 2D rects it could mean being outside towards -x, or -y or -(x, y) (current), or any of those. `meets(a, b)` could be:
```
--.--
a | b
--'--

or

| b |
|---|
| a |

or (current)

  | b
--+---
a |
```
or other rotations. Also, for the edge versions, do widths need to match for the dimension that doesn't meet?

This doesn't seem clear enough to have. It's been there since the dawn of [GeometryTypes 11 years ago](https://github.com/JuliaGeometry/GeometryTypes.jl/blame/03272fd87059bad8bc3216de82a0c40034081923/src/HyperRectangles.jl)

src/primitives/rectangles.jl
```diff
@@ -320,41 +320,74 @@ end
 # set operations
 
 """
-    isempty(h::Rect)
+    isempty(h::Rect[, volumetric = true])
```
SimonDanisch: since this is a base overload and not really self explanatory, i think this should be a kwarg.

PR #277, opened 2026-04-21
add `coordinates` for `Rect1`

thchr:
I have wanted to use the `Rect` class here occasionally in some more generic contexts, and the absence of a `coordinates` method for the 1D case has then been a bit annoying: this just proposes adding it (and adds `texturecoordinates` alongside).

There is of course more that could be done for fleshing out the `Rect{1}` methods - and for `Rect{N}` interfaces more broadly, but my ambition is limited to just this.

### DIFF
--- a/src/primitives/rectangles.jl
+++ b/src/primitives/rectangles.jl
@@ -532,8 +532,6 @@ function coordinates(rect::Rect{1, T}) where {T}
     return [Point{1,T}(o[1]), Point{1,T}(o[1]+w[1])]
 end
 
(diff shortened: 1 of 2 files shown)


ffreyer:
I'm not sure if it makes sense to define `texturecoordinates` because I don't think they make much sense outside a mesh and generating a mesh from a 1D Rect seems odd to me. If you do add them they should be using floats not ints because they are usually interpolated between vertices to sample another resource (texture).

thchr:
Makes sense: removed now.

PR #280, opened 2026-06-12
Less method invalidations for OffsetInteger

SimonDanisch:
Back in the day, I didn't have enough time to figure out why OffsetIntegers invalidate so many methods and how to fix it. The answer is, just let promotion do the hard work, and dont define unused methods :D

### DIFF
--- a/src/offsetintegers.jl
+++ b/src/offsetintegers.jl
@@ -67,16 +67,34 @@ for op in (:(+), :(-), :(*), :(/), :div)
     end
 end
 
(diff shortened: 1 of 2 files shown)


ffreyer:
This made me check if OffsetIntegers are tested and they are, but only with an offset of 0. I guess that just covers the first method and not both?
Either way, it would probably be good to extend the test to include some nonzero offset
https://github.com/JuliaGeometry/GeometryBasics.jl/blob/fcb4edfc0f2b43fdabcbe61a42ed2cf3db69e153/test/runtests.jl#L313-L341

PR #282, opened 2026-08-09
fix: avoid type var in Mesh constructor to enable JuliaC trimming

rnwst:
The type variable in the `Mesh` constructor was preventing JuliaC from being able to statically resolve types for trimming. Only use a type variable when the deprecated `normals` attribute is provided. Also fix the attribute name substitution! The previous logic had a bug in it that resulted in the code not actually substituting `normals` for `normal`.

### DIFF
--- a/src/basic_types.jl
+++ b/src/basic_types.jl
@@ -509,7 +509,7 @@ struct Mesh{
         Dim, T <: Real,
         FT <: AbstractFace,
         Names,
-        VAT <: Tuple{<: AbstractVector{Point{Dim, T}}, Vararg{VertexAttributeType}},
+        VAT <: Tuple{<:AbstractVector{Point{Dim, T}}, Vararg{VertexAttributeType}},
         FVT <: AbstractVector{FT}
     } <: AbstractMesh{Dim, T}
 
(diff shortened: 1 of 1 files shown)


src/basic_types.jl
```diff
@@ -527,17 +527,18 @@ struct Mesh{
...
         # verify type
-        if !haskey(va, :position )
+        if !haskey(va, :position)
             error("Vertex attributes must have a :position attribute.")
         end
 
         if haskey(va, :normals)
             @warn "`normals` as a vertex attribute name has been deprecated in favor of `normal` to bring it in line with mesh.position and mesh.uv"
-            names = ntuple(i -> ifelse(names[i] == :normal, :normal, names[i]), length(names))
+            names = Names
```
asinghvi17: Why is this line there?  `names` gets overridden in the next line down.  can we call this something else or just refer to `Names` directly in the comprehension?

Edit: Oh, I see this was carried over from older code.  Still, would prefer to refer to `Names` in the ntuple call then.

rnwst: My bad, I've removed that line now.

rnwst:
@asinghvi17 - ready to be merged?

SimonDanisch:
thanks!

